PR #9, opened 2021-03-01
Translate fix

seelker:
Fix French name accordingly to the original names


seelker:
Also fixed some translation swap between English and Spanish.

arnaudgregoire:
thanks for your help, i was wondering why some people reported some spanish translation in the english version.

PR #23, opened 2022-01-11
Txff

chrislayco:
f9301c3 - see commit for details

730988f - reference update, use FIREBASE_CONFIG instead

ignore 59e3dab and 90cedfd, those are only for me


arnaudgregoire:
Changes cherry picked and integrated, thx for the improvement

PR #258, opened 2023-01-11
Cache worker

sylvainpolletvillard:
Add a service worker to cache all assets and drastically improve game loading times after downloaded once. It still takes around 15 seconds on my machine to load due to the high number of requests (~1500 requests, so around 10ms per request on average).

This should also save a lot of bandwith for the server 😅 

To invalidate cache and force users to redownload the assets, you will have to update `CACHE_NAME` variable and users will have to refresh the page once. Note: I can add an update notification or force the page refresh but I believe this is not good for the user experience, and the modifications done on assets should not be that crucial that we need to force people to update. I decided not to link this cache name to the current package version because I believe most of the future updates should not require invalidating assets cache.

### DIFF
--- a/app/public/src/pages/component/game/game-pokemon-detail.tsx
+++ b/app/public/src/pages/component/game/game-pokemon-detail.tsx
@@ -37,7 +37,7 @@ export function GamePokemonDetail(props: { pokemon: Pokemon }) {
           )}
         />
         <div>
-          {props.pokemon.types.map((type) => <SynergyIcon type={type} />)}
+          {props.pokemon.types.map((type) => <SynergyIcon type={type} key={type} />)}
         </div>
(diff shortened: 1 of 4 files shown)


app/public/src/pages/component/game/game-pokemon-detail.tsx
```diff
@@ -37,7 +37,7 @@ export function GamePokemonDetail(props: { pokemon: Pokemon }) {
           )}
         />
         <div>
-          {props.pokemon.types.map((type) => <SynergyIcon type={type} />)}
+          {props.pokemon.types.map((type) => <SynergyIcon type={type} key={type} />)}
```
sylvainpolletvillard, line 40: this is not related to current PR, but to fix a warning brought by previous PR

PR #272, opened 2023-01-13
Evolution hints

sylvainpolletvillard:
Add holo effects on shop when an evolution is possible

Video: https://cdn.discordapp.com/attachments/1063164968721190974/1063288076593545327/2023-01-13_03-45-47.mp4

Also fix a client-side bug with the board not being correctly updated when removing a pokemon (see comments in changes panel)


app/public/src/game/components/board-manager.ts
```diff
@@ -68,6 +69,7 @@ export default class BoardManager {
     if (pokemonUI) {
       pokemonUI.destroy(true)
     }
+    this.pokemons.delete(pokemonToRemove.id)
```
sylvainpolletvillard, line 72: Bugfix: the board map was not cleared when a pokemon was removed. This led to ghost pokemons being kept in the board map whenever they are removed, when changing the player observed or when evolving a pokemon. @keldaan-ag  you should check again but I'm pretty sure this was not intended and could be responsible of other bugs I'm not aware of.

PR #337, opened 2023-01-29
redesign item tooltips

sylvainpolletvillard:
Redesigned item tooltips
Changes:
- now show basic item stats on advanced components
- allow to show item tooltips on item combinations
- prevent multiple item tooltips to be open at the same time
- added localization for pokemon statistics
- moved base stats of basic items in Config file

image

Remaining issues:
- for some items, it's not clear if basic stats are counted in the item description or not. For example, is Oran Bay +100HP total or +130 HP total ?
 
image

- HP bonus stats do not give HP but shield instead. This lead to not being visible in the HP stats of the pokemon, and also can confuse the players regarding to the shield/heal mechanics (notably for Monster comp). If this is intentional design, we should change the item description


sylvainpolletvillard:
This is the clarification of items having additional stats to the basic components stats:
image
image
image
image
image
image
image

PR #430, opened 2023-02-27
fix damage numbers and float number precision

sylvainpolletvillard:
Sometimes the damage number shown is 30.0000000000004
This is due to float number precision in JS:
image

calling .toFixed(0) instead of toString() fixes the issue


keldaan-ag:
sneaky !

PR #478, opened 2023-03-14
Change ability cast strategy

sylvainpolletvillard:
Initially I wanted to fix Blast Burn not touching any enemy, as discussed here: https://discord.com/channels/737230355039387749/1077020083567738931

But I think this should be fixed globally so that pokemons can cast their abilities only when they are in the same conditions than being able to basic attack.

With these changes, pokemons cast abilities only when they have a target at range. This also have these consequences:
- they won't cast abilities if frozen or asleep
- they won't do a basic attack in the same loop they cast their ability

I think this makes the behavior much more predictable and should prevent the frustration when the ability does not hit anything. Not sure how it would impact the meta, ability spammers will do a bit less basic attacks but at the same time have the guarantee to hit with their ability.

### DIFF
--- a/app/core/attacking-state.ts
+++ b/app/core/attacking-state.ts
@@ -17,7 +17,7 @@ export default class AttackingState extends PokemonState {
 
     if (pokemon.cooldown <= 0) {
       pokemon.cooldown = pokemon.getAttackDelay()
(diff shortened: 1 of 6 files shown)


keldaan-ag:
previous merged PR added some conflicts

PR #482, opened 2023-03-18
[Draft] Complete stat review

sylvainpolletvillard:
Based on the work of throat goat, I worked on formulas to evaluate a score for offensive power and defensive power for each pokemon, then defined objectives of power score based on the category, the current stage and the number of stages of the mon.

Here is the spreadsheet: https://docs.google.com/spreadsheets/d/15pTu5lNzULyg6ON52ogqubO0hCoXkBbdGNSNRSyswkc/edit?usp=sharing

I used this to make a full review of all pokemon stats to match these objectives. I have confidence in my formulas because it highlighted some mons that we already knew were problematic.

The results is mostly buffs in HP and attack. As a result, I also slightly adjusted some synergy effects in consequence. The mythicals at stage 10 and the common tier 3 are also less strong than before, compared to the upper categories ; which I think is a good thing because they were too strong in the meta

Because almost every pokemon has been modified in some way, we should merge this on a new major patch, because it is more than likely that the meta will change a lot with this, and synergies will need to be readjusted after that.


app/types/strings/Ability.ts
```diff
@@ -610,6 +610,12 @@ export const AbilityName: { [key in Ability]: Langage } = {
     prt: ``,
     fra: ``
   },
+  [Ability.JUDGEMENT]: {
```
sylvainpolletvillard, line 613: I split many mythical abilities into their own unique abilities so that it no longer relies on the "stars" property, because this property was also used for the CSV export to identify the expected strength of this mon, and therefore at which stage it should be proposed

PR #484, opened 2023-03-18
Item rework: Power Lens, Star dust, Amulet coin

sylvainpolletvillard:
Item rework to replace Bright Powder and Water Incense as discussed here:
https://discord.com/channels/737230355039387749/1084945294430576691
https://discord.com/channels/737230355039387749/1084935823117533194

Miracle Seed + Mystic Water → Star Dust: After casting ability, gain 50% of Max Mana as Shield
Twisted Spoon + Never Melted Ice → Power Lens: 50% of received spell damage is reflected to the attacker 

Also contains a fix for the visual bug of shield remaining in the life bar when at zero


app/public/src/game/components/battle-manager.ts
```diff
@@ -427,7 +427,7 @@ export default class BattleManager {
               pkm.detail.hp.textContent = pokemon.life.toString()
             }
           } else if (change.field == "shield") {
-            if (change.value > 0) {
+            if (change.value >= 0) {
```
sylvainpolletvillard, line 430: this is the fix for the visual bug of shield remaining visible in life bars when at zero

PR #633, opened 2023-05-17
Allow to refill empty slots of the shop when locked

elias-bina:
Allow the empty slots of the shop to be refreshed even when it is locked

I think that the lock could then stay locked for the next round, but since it is debatable I didn't changed it


elias-bina:
I have checked the functionnality and the fact that ditto still appears

elias-bina:
Added a fix for ditto (which was eating the pool), also tested

PR #634, opened 2023-05-18
fix(ditto): Prevents loss of object when duplication

elias-bina:
This is a fast way to solve the problem

I think will try to create an "addItem" function on pokemons twhich will take the items and the player, wich will perform all the chevk necessary. 

This could make a common way to handle this case, the evolution case and the drop item all in one, since for now evolution only copies fused objects


elias-bina:
(Fixes [This](https://discord.com/channels/737230355039387749/1108704055674544159) )

PR #638, opened 2023-05-20
fix(dragdrop): Prevents neutral pokes from going on board

elias-bina:
Eggs were able to be put on board (only by human since `getFirstPokemonOnBench` didn't take them), so prevented any NEUTRAL pokemon from being set on board by `OnDragDropCommand`

Removed setting `updateBoard` to false on success since it doesn't trigger the `DRAG_DROP_FAILED`

### DIFF
--- a/app/rooms/commands/game-commands.ts
+++ b/app/rooms/commands/game-commands.ts
@@ -243,25 +243,28 @@ export class OnDragDropCommand extends Command<
             success = true
           }
         } else {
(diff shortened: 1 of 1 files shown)


elias-bina:
Tested on the 8 cases of `(!isBoardFull || !dropToEmptyPlace || !dropFromBench)`, and with / without Egg (with egg could obviously test only dropFromBench lol)

PR #706, opened 2023-06-06
compute streak and round damage before increasing stage level

sylvainpolletvillard:
fix for https://discord.com/channels/737230355039387749/1115772606218502205

another consequence of the change to stage number computation, this cause bug in streak computation


sylvainpolletvillard:
i let you decide if you want to hotfix or not

PR #718, opened 2023-06-10
Ghost damage

David-Helmerson:
Fixed multiple issues relating to ghost synergy damage partitioning
- True damage from ghost synergy was omitted in damage calculation for choice scarf
- Full true damage from ghost 8 would result in the attacker not gaining mana from the basic attack


app/core/attacking-state.ts
```diff
@@ -326,15 +304,28 @@ export default class AttackingState extends PokemonState {
...
+              board,
+              attackType,
+              attacker: pokemon,
+              dodgeable: true,
+              shouldAttackerGainMana: false,
+              shouldTargetGainMana: true
+            })
+
+            // Apply ghost true damage
             cell.value.handleDamage({
```
sylvainpolletvillard, line 319: missing another if (ghostTrueDamage > 0) { ?

PR #727, opened 2023-06-11
onHit effects classification

sylvainpolletvillard:
There's a lot of confusion and bugs around on-hit effects, so I refactored it to make a clear distinction between the 3 categories of effects i saw:

1) onHit : whenever the unit does a successful basic attack (not dodged or protected)
2) onAttack: whenever the unit do a basic attack, no matter if it's successful or not
3) onDamageDealt: whenever the unit deals damage, by attack or ability

I put in bold the things that this PR has modified

## onHit effects

Many on-hit effects now only apply when the attack is successful (not dodged or protected). This concerns:

- **Upgrade stacks**
- **Aquatic mana burn**
- **Fire attack buff**
- **Poison chance from Poison synergy**
- **Freeze chance from Ice synergy**
- **Silence chance from Ghost synergy**
- **Red orb additional damage**
- **Smoke Ball debuff**
- **Spike Armor**

## onAttack effects

The effects that still apply no matter if the attack is successful or not :
- Dragon attack speed buff
- Choice Scarf
- Leftovers
- Mana Scarf
- Delta Orb
- Blue Orb
- Defensive Ribbon stacks
- Shiny charm protect
- **Mana gain on attack**

## onDamageDealt effects

- Human heal
- possibily others ?

everything in this list can be discussed ofc

# Additional bugfixes in the PR:
- Leftovers heal was triggering twice for the holder


sylvainpolletvillard:
Requesting review from @David-Helmerson , since you also worked on damage computation recently

David-Helmerson:
I like this distinction for effects, however the majority of issues I've run into are in relation to the behavior of handleDamage. Either all behavior in the function call should be distributive across multiple calls within the same attack, or handleDamage should handle all portions of damage from a single source at once before the target's HP is reduced.

Currently there are several non-distributive effects within handleDamage that I'm not certain fit into any of the proposed categories:
- Fighting synergy damage reduction
- Flying synergy protection (could be onDamageDealt)
- Fossil synergy (without any sort of revive delay)
- Fire gem double shield damage
- Fire gem 0.1 * hp damage (could be onHit like red orb)
- Wonder guard (Maybe could be onHit???)

---
To put forward an idea of how an attack could be executed:
1. Attack is made
2. All onAttack effects are applied
3. Hit/miss is calculated
4. Crit is calculated
5. onCritical effects are applied
7. onHit effects are applied, which in part calculate a single partitioned damage total
8. All damage to target is calculated
9. onDamageDealt effects are applied (twice for choice scarf), which could include the actual reduction of target HP
10. Potentially a separate onDeath for neatness

I've tried to think about most cases, and ultimately my take away is that a handleDamage refactor should come alongside these changes.

sylvainpolletvillard:
Looks good to me !

Regarding the remaining effects in handleDamage, here are my thoughts:

Fighting synergy damage reduction 
→ should be moved to attack(), i'm not even sure we want it to block special damage in the first place

Flying synergy protection (could be onDamageDealt) 
→ can be triggered in handleDamage, doesn't change the behavior

Fossil synergy (without any sort of revive delay) 
→ can be triggered in handleDamage, doesn't change the behavior

Fire gem double shield damage 
→ can be triggered in handleDamage, it slightly changes the behavior but i think it's okay

Fire gem 0.1 * hp damage  
→ should be moved to attack() on physicalDamage calculation

Wonder guard 
→ i like it in handleDamage too, would make red orb a counter item

sylvainpolletvillard:
I agree separate onDeath effects would be cleaner, but let's do it step by step.

I'm removing draft status if it's okay for you to merge

PR #797, opened 2023-07-08
fixed POKE_DOLL, electricField, psychicField, grassField, fairyField,…

TBehmann:
… FIRE_GEM and MISTY not working correctly

With this fix the amplification of the items, fields and weather will also lead to more or less mana gain. For the item poke doll this behavior is maybe debatable.


sylvainpolletvillard:
DO NOT MERGE, this is not working and actually breaks all the mechanis mentionned

PR #828, opened 2023-07-20
changed FLUFFY_TAIL stats from SPE_DEF to DEF

TBehmann:
Fluffy tail is giving spe def instead of def in the moment. Considering change notes and the build path def is probably the intended stat. This PR fixes the problem.


sylvainpolletvillard:
LGTM

PR #864, opened 2023-08-06
use shiny pokemon if user has shiny pokemon avatar

campavao:
for the user's pokemon avatar that appears in game, use shiny pokemon if user's avatar is shiny


campavao:
Example in the carousel item picker:
image

PR #865, opened 2023-08-06
show pokemon/item tooltip on hover instead of click

campavao:
as title suggests, change when tooltips appear. This helps fix when tooltips are persistent on screen until you reclick that same pokemon.

hopefully not overstepping my bounds here, I think this could be a nice QoL improvement 

https://github.com/keldaanCommunity/pokemonAutoChess/assets/25358150/61301ab0-4a79-4a2a-aae4-37bef206999f


keldaan-ag:
item tooltip needs to stay on click. Ppl can hover tier2 items in the tier1 item description.

PR #868, opened 2023-08-06
wip persist item details

campavao:


sylvainpolletvillard:
not sure but i think this one can be closed too after https://github.com/keldaanCommunity/pokemonAutoChess/pull/922
@campavao can you confirm ?

PR #869, opened 2023-08-06
Sidebar + Responsive Lobby

campavao:
Adds `MainSidebar` which now houses the menu buttons.
- sidebar is collapsed on default
- sidebar opens/closes on mouse enter/leave
- most options in sidebar now open in modals instead of pages

Page -> Modal components:
- collection
- boosters
- meta
- wiki
- news
- options

News moved from lobby screen into modal

Sidebar now shown on Preparation room
- now allows users to check out things like the meta and wiki while they wait for a game to start

General responsiveness improvements to lobby and other modals
- online + chat window flex to a column at smaller breakpoints

### Problems
Booster animation is a little different than how it was before. I don't think it's awful, but probs worth updating again in a separate PR


sylvainpolletvillard:
Great work ! Few bugs/remarks:

image
checkbox vertical align

image
scrollbar on tabs list

image
white square at bottom right

image

same white square on many modals in the wiki

image
bot admin needs to be replaced with i18n label (maybe fixed by merging last master)

image
the icon is a bit weird, may i suggest https://game-icons.net/1x1/delapouite/exit-door.html

image
wiki modal content could use a max height with internal scrollbar, so that the tabs list of the wiki stays on top

sylvainpolletvillard:
on larger resolutions, could use some space available left, maybe for chat width ?
image

sylvainpolletvillard:
also probably not very important, but the new layout is not usable for viewport width < 1000px (the sidebar disappears). In the previous layout, i made it so the blocks go vertical in that case. Even if the game is not suitable for mobile or portrait views, I think we should make it usable at least, especially since this PR is about making the lobby responsive.

campavao:
> also probably not very important, but the new layout is not usable for viewport width < 1000px (the sidebar disappears). In the previous layout, i made it so the blocks go vertical in that case. Even if the game is not suitable for mobile or portrait views, I think we should make it usable at least, especially since this PR is about making the lobby responsive.

ah that's just an artifact from having `breakPoint` on the sidebar. If I remove that it persists for mobile

sylvainpolletvillard:
i fixed the boosters anim here: https://github.com/campavao/pokemonAutoChess/pull/1/

PR #917, opened 2023-08-17
hide FAQ and How to play videos in game

campavao:


app/public/src/pages/main-sidebar.tsx
```diff
@@ -42,12 +42,10 @@ export function MainSidebar(props: MainSidebarProps) {
   const {
     meta,
     metaItems,
-    user,
-    botList
```
campavao, line 46: this wasn't being used so I removed it

PR #919, opened 2023-08-17
add basic sprite debug page

campavao:
creates admin/mod only sprite debug page

currently allows for picking from any pokemon and any status, only in forward facing position

also adds the package [react-bootstrap-typeahead](https://www.npmjs.com/package/react-bootstrap-typeahead)

future updates:
- increase scale
- change pokemon direction
- change shiny
- more???

https://github.com/keldaanCommunity/pokemonAutoChess/assets/25358150/3b6b2f0d-7a8d-480f-9ff6-78aa4b49d50b

### DIFF
--- a/app/public/src/pages/component/lobby-menu/sprite-debug.css
+++ b/app/public/src/pages/component/lobby-menu/sprite-debug.css
@@ -1,0 +1,51 @@
+.sprite-debug-root {
+    width: 100%;
+    height: 100%;
+
+    --sidebar-offset: 90px;
+}
+
+/* TODO Remove this after the absolute sidebar update goes in */
+.sprite-debug-root aside.sidebar {
+    position: absolute;
+    height: 100%;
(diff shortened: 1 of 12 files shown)


app/public/src/pages/component/lobby-menu/sprite-debug.css
```diff
@@ -0,0 +1,51 @@
+.sprite-debug-root {
+    width: 100%;
+    height: 100%;
+
+    --sidebar-offset: 90px;
+}
+
+/* TODO Remove this after the absolute sidebar update goes in */
+.sprite-debug-root aside.sidebar {
```
campavao, line 9: This update will be provided in `.main-sidebar.css` with [this PR](https://github.com/keldaanCommunity/pokemonAutoChess/pull/918)  but I kept this fallback in case we don't end up merging that one

PR #941, opened 2023-08-25
fix game player detail synergies not being accurate

campavao:
I think there was a slight regression in the 1v1 update where this was removed.

Re-added this logic + updated key 

Let me know if you removed this intentionally, if so we should probably hide synergies in general here as they're not updated properly without this.

Saying that, this really updates after a round so it's still not real time, but as close to it as we can get for now without having to add in a bunch more logic for updating game state synergies.


sylvainpolletvillard:
probably a merge conflict, not intentional removing. Happened to me on a few PR as well

keldaan-ag:
I don't want network to be used in a selector. It makes the component render everytime, with a big impact on performance. I also know that theres a problem with synergies not up to date.

PR #954, opened 2023-08-27
Clean up game.tsx for better perf

campavao:
- General clean up of warnings and errors
  - all logic outside of `game.tsx` has not been modified, just removed unused params/imports and updated types (and/or removed `any`s)
  
- Clean up of `game.tsx`
  - converted functions into useCallbacks (for dependencies)
  - updated connectToGame so that it doesn't attempt to reconnect while still connecting 

Overall I think the game is running smoother due to less unnecessary re-renders, but we should test this one heavily since it touches the connect/reconnect to game logic 

I tested locally and am seeing it work properly (can get into game, uses first attempt to do so) but again encourage more testing from y'all to make sure we square away any bugs.


app/public/src/pages/game.tsx
```diff
@@ -127,47 +124,52 @@ export default function Game() {
...
-    const cachedReconnectionToken = localStorage.getItem(
-      "cachedReconnectionToken"
-    )
-    if (cachedReconnectionToken) {
-      connecting.current = true
-      const statusMessage = document.querySelector("#status-message")
-      if (statusMessage) {
-        statusMessage.textContent = `Connecting to game...`
-      }
+  const connectToGame = useCallback(
```
campavao, line 127: diff looks scary here, but only changes to this function were to convert to `useCallback` and adding async/await on line 155

PR #966, opened 2023-08-29
redesign online avatars

campavao:
image

wanted to blow up the user avatar image a bit and squish the height of the overall container to fit more players into the online column

could extend these in the future for other areas where we show users


sylvainpolletvillard:
image


It feels weird that the player name is smaller than his title. Also on some resolutions, the image does not cover full height

image

PR #1000, opened 2023-09-06
reduce flora/ice portal symbols count

sylvainpolletvillard:
make Ice 2, Ice 3 and Flora 3 no longer give a Portal symbol, because they were overrepresented due to the higher number of synergy triggers


sylvainpolletvillard:
PR 1000 😎

PR #1061, opened 2023-10-01
Fix ghost income

sylvainpolletvillard:
Fix two problems

1) money was stored in UINT8 meaning it resets to 0 when going over 256. Changed to UINT16

https://github.com/keldaanCommunity/pokemonAutoChess/assets/566536/98a0978d-6466-482f-949a-d230b31776f9

2) Dereference ghost player when assigning matchups so that it does not apply bonus income (Amulet coin, payday etc...) to original player. Should work as well for pokemon gain or item gain with Unowns hidden power


sylvainpolletvillard:
I've merged to prod branch but up to you @keldaan-ag  if you want to make this a hotfix or not.

We will need to merge prod on master anyway

PR #1234, opened 2023-11-15
pokemon release fixes

David-Helmerson:
Least important bugs ever fixed

- Selling Alolan Raichu (4 star unit) would void 9 Pichus from the shop pool
- Selling Ninjask + Shedinja would create a surplus of Nincada in shop pool, and return more gold than the buy price of constituent Nincada
- Changed Shedinja stars to 0 for consistency in solving above, should have no adverse effects besides Roar of Time interaction

### DIFF
--- a/app/models/colyseus-models/pokemon.ts
+++ b/app/models/colyseus-models/pokemon.ts
@@ -6565,7 +6565,7 @@ export class Ninjask extends Pokemon {
 export class Shedinja extends Pokemon {
   types = new SetSchema<Synergy>([Synergy.BUG, Synergy.GHOST, Synergy.FLYING])
   rarity = Rarity.EPIC
-  stars = 2
+  stars = 0
   hp = 10
   atk = 18
   def = 0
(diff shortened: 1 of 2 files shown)


app/models/colyseus-models/pokemon.ts
```diff
@@ -6565,7 +6565,7 @@ export class Ninjask extends Pokemon {
 export class Shedinja extends Pokemon {
   types = new SetSchema<Synergy>([Synergy.BUG, Synergy.GHOST, Synergy.FLYING])
   rarity = Rarity.EPIC
-  stars = 2
+  stars = 0
```
sylvainpolletvillard, line 6568: this is not the correct way to solve the issue, Shedinja is a tier 2

if you want to change sell price of Shedinja, do it in getSellPrice() function

We could also fix the issue by considering Shedinja/Munja a Pokemon duo. We already have special rules for sell prices of unique duos

sylvainpolletvillard:
since no update for a week, I fixed it myself in #1267

PR #1284, opened 2023-12-02
set Eeveelutions family to Eevee

sylvainpolletvillard:
Eevee can be found in shops after evolving another Eevee. I'm assuming this is a bug ?


sylvainpolletvillard:
Apparently players like it this way

PR #1294, opened 2023-12-07
small rework of the attack method to fix crits bypassing protect/ dodge and Xurkitree crits being physical

TBehmann:
In the current code a crits bypasses dodge and protect. Additionally, if the damage is converted to special damage, crits will still be physical. Which is counterintuitive at least. True damage also does only work with physical damage. This PR tries to fix these problems with a small rework of the attack method.

### DIFF
--- a/app/core/attacking-state.ts
+++ b/app/core/attacking-state.ts
@@ -121,50 +121,12 @@ export default class AttackingState extends PokemonState {
         pokemon,
         target
       )
-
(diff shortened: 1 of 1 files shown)


sylvainpolletvillard:
small bug with Spot panda additional damage but besides that, looks good to me. Thanks !

sylvainpolletvillard:
I'll take care of it

PR #1381, opened 2024-01-08
fix deleting artificial items can delete synergies from units not wearing the artificial item

TBehmann:
https://discord.com/channels/737230355039387749/1193687245425750046

### DIFF
--- a/app/models/colyseus-models/player.ts
+++ b/app/models/colyseus-models/player.ts
@@ -209,9 +209,11 @@ export default class Player extends Schema implements IPlayer {
 
   updateSynergies() {
     const pokemons: Pokemon[] = values(this.board)
-    const updatedSynergies = computeSynergies(pokemons)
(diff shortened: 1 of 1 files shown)


app/models/colyseus-models/player.ts
```diff
@@ -209,10 +209,12 @@ export default class Player extends Schema implements IPlayer {
 
   updateSynergies() {
     const pokemons: Pokemon[] = values(this.board)
-    const updatedSynergies = computeSynergies(pokemons)
+    var updatedSynergies = computeSynergies(pokemons)
 
     this.updateArtificialItems(updatedSynergies)
 
+    updatedSynergies = computeSynergies(pokemons)
```
sylvainpolletvillard, line 216: we should compute twice only if artificial items have been lost, to save some performance. This can be done in another PR if you want

sylvainpolletvillard:
Looks good to me, thanks 👍

PR #1382, opened 2024-01-08
fix removing artificial items were not updating the synergy list

sylvainpolletvillard:
follow up on https://github.com/keldaanCommunity/pokemonAutoChess/pull/1381

this time for prod


sylvainpolletvillard:
mb i think we don't have another choice than computing twice, closing this one

PR #1442, opened 2024-02-02
update on some spanish translations

Abogadouuu:
translated some ability names and descriptions.
Also added pokemon name translation (most of them are the same as english,but also i translated busted from mimikyu and the colors of mimikyu


sylvainpolletvillard:
Thanks for keeping up ! Appreciated <3

PR #1541, opened 2024-03-04
Update translations

KrownK:
Describe the changes you have conducted here

Preview the messages on https://inlang.com/github.com/KrownK/pokemonAutoChess .


sylvainpolletvillard:
Thanks ! Note that some labels translated here may change because they're related to work in progress for next patch

PR #1609, opened 2024-03-27
fairy splash handleSpecialDamage

David-Helmerson:
Fairy splash damage was handled by handleDamage instead of handleSpecialDamage, preventing interactions with power lens, magic bounce, and pokenomicon


sylvainpolletvillard:
I think this was on purpose, but I agree it's not clear. Usuaullt handleDamage is used instead of handleSpecialDamage to prevent infinite loop scenarios (like magic bounce vs magic bounce). We need to make sure there's no such case with fairy splashes.

sylvainpolletvillard:
I think we're good with the explicit `crit:false` flag, I don't see any infinite loop case

PR #1692, opened 2024-04-28
fix a bug causing to fish always epic Pokémon

TBehmann:
Due to a missing break only epic Pokémon could be fished. This PR fixes this issue.


sylvainpolletvillard:
good catch

PR #1724, opened 2024-05-05
Update spanish translations

matiasdelre7:
Describe the changes you have conducted here

1- The word ELEMENTOS (elements) is changed to OBJETOS (items).  The appropriate translation for ITEMS is OBJETOS as it is better understood and is the one used in all pokemon games.
2-Correction of upper letters in some bad translations.
3-Translation of some Pokemon-skills from the new patches is added.


sylvainpolletvillard:
Thank you !

PR #1734, opened 2024-05-09
precompress with brotli static assets after assetpack

sylvainpolletvillard:
Replace on-the-fly response gzip compression with brotli pre-compression for assets and static files

This should reduce significantly the server CPU load while optimizing further the assets

observed results: comparison with prod (gzip → brotli): 36.9MB → 33.9 MB (not a fair comparison because more assets have been added in the meantime, so the gain is probably more)

## Drawback

brotli compression takes around 3 minutes on my high-end computer on 16 threads. It's probably too much for continuous integration / postinstall on server

If we use an external CDN service, all of this is not necessary. But if not, we may have to consider commiting compressed files (which i hate because repo is already bloated, will make inlang even slower etc.)  or maybe we could have a repo dedicated to post-build compressed assets

chrome_hWHwdDGSTc
Discord_2urqHEWwI0


sylvainpolletvillard:
closing in favor of Cloudflare autocompression

PR #1747, opened 2024-05-12
trace room owner name for room name moderation

sylvainpolletvillard:
There's a few things we can do for better moderation of room names. Renaming room when room leader changes, and showing current room leader in tooltip


keldaan-ag:
best picture example !

PR #1770, opened 2024-05-18
Update translations

lexusbest1:
Added some spanish translations

Preview the messages on https://inlang.com/github.com/lexusbest1/pokemonAutoChess .


sylvainpolletvillard:
keywords in uppercase like POISONNED or SHIELD should not be translated

PR #1775, opened 2024-05-19
Fixed Comfey being duplicated when given to flora spawn

sylvainpolletvillard:
A really funny proc with flora is comfey, if a flora mon dies with comfey on it and moves to the gloom it'll proc again when it dies and duplicates comfey

https://discord.com/channels/737230355039387749/737244224688226346/1241811092733165628


keldaan-ag:
👍

PR #1776, opened 2024-05-19
fix comfey on pokemons that cant hold items

sylvainpolletvillard:
- Comfey can no longer be hold by pokemons that cannot hold items like Unowns
- Ditto can now hold items. Items are dropped when Ditto is transforming to another pokemon.


keldaan-ag:
👍

PR #1788, opened 2024-05-22
Trubbish

Windchili:
Added Trubbish and Garbodor


sylvainpolletvillard:
Some conflicts from recent merge were not fixed correctly, i let you check Trubbish aftersimulationStart for typescript errors

sylvainpolletvillard:
i fixed them so that i can test

sylvainpolletvillard:
Okay looks good to me. I think buffing to 25 shield per item component would be closer to the right value

Windchili:
Thanks for fixing those.
I agree its not really worth it to give garbodor item components. I'm happy to give 25 instead of 10, just checking you meant to say shield and not HP.
i.e.
Item components: +25 Shield and Item stats
Berries: +10 HP
Artificial Items: +50HP, Item stats and Trash

sylvainpolletvillard:
Yeah 25HP

sylvainpolletvillard:
good to merge

PR #1829, opened 2024-06-07
Fix decimal stats

DevGolden:
Fix previous correction + add new corrections into crit chance and attack speed

### DIFF
--- a/app/core/pokemon-entity.ts
+++ b/app/core/pokemon-entity.ts
@@ -365,9 +365,7 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
     apBoost: number,
     crit: boolean
   ) {
-    value = Math.round(
-      value * (1 + (apBoost * caster.ap) / 100) * (crit ? caster.critPower : 1)
-    )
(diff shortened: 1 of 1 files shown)


app/core/pokemon-entity.ts
```diff
@@ -365,9 +365,7 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
     apBoost: number,
     crit: boolean
   ) {
-    value = Math.round(
-      value * (1 + (apBoost * caster.ap) / 100) * (crit ? caster.critPower : 1)
-    )
+    value = value * (1 + (apBoost * caster.ap) / 100) * (crit ? caster.critPower : 1)
```
sylvainpolletvillard, line 368: critChance is a value between 0 and 100, so rounding it is fine

app/core/pokemon-entity.ts
```diff
@@ -461,9 +459,7 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
...
```
sylvainpolletvillard, line 462: ability power can be rounded as well

PR #1859, opened 2024-06-19
improve spanish translations

matiasdelre7:
:)


matiasdelre7:
the conflict is that dynamax band has 2 descriptions, but I can't fix it.

PR #1864, opened 2024-06-20
Move Wiki into Sidebar

campavao:
**Ok definitely pull this down and review this before merging.** 

With this PR I updated the Wiki to open within the Sidebar rather than in a modal. Idea is two-fold:
1. Can now keep the sidebar wiki pinned open during game
2. Wiki is now mobile responsive

I wanted to make this update because there are many times where I want to search something during the match and the wiki modal can sometimes take too long to load while blocking my entire screen. 

Some other updates bundled with this:
- Pokemon section no longer uses Tabs and instead uses `select` dropdowns
- Pokemon section also has been cleaned up a bit
  - Now when viewing a single pokemon, you no longer see duplicate information 
  - When viewing a single pokemon, the rest of the list of pokemon in that category disappears and is replaced by a clear button
- Abilities section now has a search bar where you can search the ability name or a pokemon that has that ability. This search is debounced to help performance while typing

Other sections have had images scaled down. Future improvements I'd like to make are:
- give Regions, Weather, and Synergies a search bar to search similarly to Abilities section
- update Synergies to get rid of Tabs and use `select`

I'm not totally sure, but I feel like Tabs might be resulting in some performance issues across the board. I know it's used widely and I don't totally have concrete proof other than the Pokemons section seems to be a bit faster now without it.

I did leave a lot of the old code for the modal wiki. I think in areas like the lobby it might make sense to use the modal, but I do like the aspect of being able to interact with the chat and join a game quickly with the wiki opened.

https://github.com/keldaanCommunity/pokemonAutoChess/assets/25358150/aceceaef-b50c-4223-ae0a-0b2ee1343131

https://github.com/keldaanCommunity/pokemonAutoChess/assets/25358150/0f24e4c7-7035-4700-9997-3b3d24744daa

https://github.com/keldaanCommunity/pokemonAutoChess/assets/25358150/60573ee7-4066-457e-9011-0023b66196f6

https://github.com/keldaanCommunity/pokemonAutoChess/assets/25358150/e75994cf-e228-4b0e-b038-30898bb07755

### DIFF
--- a/app/public/src/pages/component/main-sidebar/main-sidebar.css
+++ b/app/public/src/pages/component/main-sidebar/main-sidebar.css
@@ -3,6 +3,12 @@ aside.sidebar {
   height: 100%;
   z-index: 100;
   background-color: var(--color-bg-primary);
+  max-width: 20vw;
+}
+
+aside.sidebar-wide {
+  width: 100%;
+  min-width: 340px;
 }
 
 .sidebar-logo {
@@ -64,7 +70,7 @@ aside.sidebar {
 }
 
 .ps-menu-root {
-  height: 100%;
+  min-height: 100%;
   background-color: var(--color-bg-primary);
 }
 
@@ -211,4 +217,16 @@ aside.sidebar {
   100% {
     opacity: 1;
   }
+}
+
+.sidebar-section {
+  position: relative;
+  min-height: 100%;
+  background-color: var(--color-bg-primary);
+}
+
+li.go-back-btn {
+  position: sticky;
+  bottom: 0;
+
 }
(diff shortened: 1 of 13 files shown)


zBrodie:
Being able to have the wiki open to the items page will be very very useful for new players. If this change is controversial, is it possible to add a full screen button that opens the old Wiki layout instead?

campavao:
> Being able to have the wiki open to the items page would be very very useful for new players. Is it possible to add a full screen button that opens the old Wiki layout?

Glad you think so! I was planning on supporting the old way as well but just haven't found the time yet. I'll try to come back to this though if there's some interest

PR #1897, opened 2024-07-04
Incense and Electirizer rework

sylvainpolletvillard:
- Rework Electirizer: Every third attack burns 15 PP of the target and inflicts PARALYSIS for 2 seconds
- Rework Incense: No longer give Flora synergy. Prevent wild Pokémon encounter in shop when held. Instead, each Pokémon in shop has an additional 5% chance to share a common synergy with the holder.


app/models/colyseus-models/player.ts
```diff
@@ -232,6 +231,14 @@ export default class Player extends Schema implements IPlayer {
       updatedSynergies = computeSynergies(pokemons)
     }
 
+    const rockNeedsRecomputing = this.updateWeatherRocks(updatedSynergies)
+    if (rockNeedsRecomputing) {
+      /* NOTE: in some edge cases like losing Hard Rock artif item, we may need to compute synergies
```
sylvainpolletvillard, line 236: this fixes https://discord.com/channels/737230355039387749/1258187979810279524

PR #1904, opened 2024-07-08
Klefki + Locked Status

Windchili:
New status: Locked: Prevents moving and sets Attack Range to 1.
(Attack range change was added as ranged attackers were rarely hindered by the status)

Klefki's ability went through many revisions. Just locking a unit for X seconds is too inconsistent, and AP scaling lock can be risky. The final version of Fairy Lock is the one I found most fun and balanced.


app/core/abilities/abilities.ts
```diff
@@ -9081,6 +9081,53 @@ export class RockHeadStrategy extends AbilityStrategy {
...
+    const cells = board.getAdjacentCells(
+      target.positionX,
+      target.positionY,
+      true
+    )
+
+    let inhabitedCells = new Array<Cell>()
+    let enemiesHit = 0
+
+    cells.forEach((cell) => {
```
sylvainpolletvillard, line 9103: can be replaced by a `.filter`

app/core/abilities/abilities.ts
```diff
@@ -9081,6 +9081,53 @@ export class RockHeadStrategy extends AbilityStrategy {
...
+          Math.round(90 / enemiesHit),
+          board,
+          AttackType.SPECIAL,
+          pokemon,
+          crit
+        )
+      }
+    })
+    
+    target.status.triggerLocked(1500 * enemiesHit, target)
```
sylvainpolletvillard, line 9127: i'm not a fan of this design, scaling with nb of enemies hit. Feels like a double penalty and drastically change the strength of the move. i would prefer a fixed delay. We learned the lesson in the past that all the hard CC status effects durations should not scale

app/models/colyseus-models/status.ts
```diff
@@ -1074,4 +1083,30 @@ export default class Status extends Schema implements IStatus {
...
+
+    ) {
+
+      if (pkm.status.enraged) {
+        duration = duration / 2
+      }
+
+      this.locked = true
+      this.lockedCooldown = Math.round(duration)
+      pkm.range = 1
```
sylvainpolletvillard, line 1100: should change pokemon state to IdleState

app/core/pokemon-entity.ts
```diff
@@ -508,10 +508,12 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
   }
 
   skydiveTo(x: number, y: number, board: Board) {
-    board.swapValue(this.positionX, this.positionY, x, y)
-    this.status.skydiving = true
-    this.toMovingState()
-    this.cooldown = 1000 // 500ms for flying up and 500ms for skydive anim
+    if (!this.status.locked){
```
sylvainpolletvillard, line 511: should be handled beforehand (see canMove usage for flying skydiving)

app/core/moving-state.ts
```diff
@@ -25,7 +25,7 @@ export default class MovingState extends PokemonState {
         pokemon,
         board
       )
-      if (pokemon.status.charm) {
+      if (pokemon.status.charm && !pokemon.status.locked) {
```
sylvainpolletvillard, line 28: should use canMove getter

app/core/moving-state.ts
```diff
@@ -47,7 +47,7 @@ export default class MovingState extends PokemonState {
           pokemon,
           board
         )
-        if (targetAtSight) {
+        if (targetAtSight && !pokemon.status.locked) {
```
sylvainpolletvillard, line 50: should use canMove getter

sylvainpolletvillard (changes requested):

app/models/colyseus-models/status.ts
Windchili: Correct me if I'm wrong. Would changing Pokemon state to IdleState make Locked functionally the same as Freeze, as the Pokemon would be unable to attack?
I've addressed all the other comments in my local repo except this one, just wanted to check before committing and pushing this

sylvainpolletvillard: right, i see we don't have a IdleState > Attacking state condition. It's probably a mistake because transitionning to moving state introduce a ~500ms delay before attacking at range. But maybe it's intentional ? Anyway, idle state is almost not used during battle so not a big deal. My intent with this remark was to not have a locked pokemon in MovingState since it obviously can't move. It avoids checking against a canMove condition at every update. But if its simpler to handle this logic in Movingstate, I can bear with that

Windchili (commented):
Addressed all requested changes, except Locked status setting Pokemon toIdleState

sylvainpolletvillard (approved):

PR #1936, opened 2024-07-21
Bugfix - Mew moved from Unique to Legendary Shop

Windchili:


Windchili:
https://discord.com/channels/737230355039387749/1264482817186857053

Mew was still in unique shop, despite now being legendary

PR #1979, opened 2024-08-01
Small eruption rework

keldaan-ag:


app/public/src/stores/GameStore.ts
```diff
@@ -122,13 +122,13 @@ export const gameSlice = createSlice({
       state.shop = action.payload.map(p=>p)
     },
     setItemsProposition: (state, action: PayloadAction<ArraySchema<Item>>) => {
-      state.itemsProposition = JSON.parse(JSON.stringify(action.payload))
+      state.itemsProposition = action.payload.map(p=>p)
```
sylvainpolletvillard, line 125: to clone array [...action.payload] is slightly faster

PR #1982, opened 2024-08-01
Cramorant

Windchili:
Cramorant + Arrokuda

To stop Pokemon from occupying Arrokuda's tile before it spawns in, I have made a new Effect.RESERVED_TILE. However it will be easy to change this code to create other tile effects that block off the tile entirely (As a new design space for tile effects in the future.)

There were quite a few edge cases that were solved, but it's possible one slipped through. I'll be on the lookout for those.


keldaan-ag:
was scared by reserved tiles  so went for target + getNearestAvailablePlaceCoordinates

PR #1987, opened 2024-08-02
Pawmi Line

Windchili:
Regional Rare Elec/Fighting
Double Shock: The user emits a powerful electric shock, dealing [50,100,200,SP] SPECIAL. The user is PARALYSIS for 3 seconds.

Implementation of Crab's design.


Windchili:
Need to fix sprites

PR #1989, opened 2024-08-03
Update translations

ThunderCapivara:
Added translations to Brazilian Portuguese

Preview the messages on https://inlang.com/github.com/ThunderCapivara/pokemonAutoChess .


sylvainpolletvillard:
@ThunderCapivara keywords in uppercase like SPECIAL, LIGHT, ATK needs to kept as is, otherwise they won't be replaced with their icons in the tooltips

PR #1993, opened 2024-08-04
Suggestion: Status.ts housekeeping + ghost curse bugfix

CostaPrimo:
This turned into a rather large pull request, sorry about that..
Amongst the changes is an attempted fix for the ghost curse bug related to status clearing / immunity.

* Extracted Aquatic Modifier to private method to reduce code duplication
* Renamed variables for clarity purposes (Cooldown vs Duration)
* Now always update durations before checking if the duration is at or below 0 rather than checking if the new duration would be at or below 0 and reducing the duration otherwise.
* Moved burn damage rounding to handleDamage like it is done with poison damage.
* Removed this.origin condition from updateBurn to make it consistent with poison.
* Changed burnOrigin, silenceOrigin, and woundOrigin to use provided origin if present or keep current origin (behavior unchanged, but reformatted into a 1-liner) 
* Added healStatus methods for negative statuses that didnt already have one.
* Reworked clearNegativeStatus to use new healStatus methods and changed uses accordingly.
* Changed triggerRuneProtect to accomodate new clearNegativeStatus parameter.
* Changed ghost curse application conditions to check for either the effect or the boolean for their respective curses in updateAllStatus method
* Removed unecessary line from LUM_BERRY related to clear status since rune protect is triggered right after resulting in the same effect.
* Removed uncessary line from updateSynchro setting this.synchro to false since it is set to true again in the very next line.
* Changed berries related to status to use the new healStatus methods.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules/
+.idea
 .vscode/
 *.env
 .env.development
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,7 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <ScalaCodeStyleSettings>
+      <option name="MULTILINE_STRING_CLOSING_QUOTES_ON_NEW_LINE" value="true" />
+    </ScalaCodeStyleSettings>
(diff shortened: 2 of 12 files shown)


CostaPrimo:
I'll try to separate this into smaller chunks rather than 1 big PR.

PR #1994, opened 2024-08-05
bugfix: Ghost curse statuses not re-applying after being cleansed

CostaPrimo:
https://discord.com/channels/737230355039387749/1267637117442461738

applyCurse only sets the curseFlag, it never adds the curse effect to list of effects on the pokemon, this the status   never reapplies if removed.

since the curseFlags are set with applyCurse, checking the appropriate flag should fix the issue.

### DIFF
--- a/app/models/colyseus-models/status.ts
+++ b/app/models/colyseus-models/status.ts
@@ -259,7 +259,7 @@ export default class Status extends Schema implements IStatus {
     }
 
     if (pokemon.curseFate && !pokemon.status.curse) {
-      this.triggerCurse(6000)
+      this.triggerCurse(5000) //Intentionally at 5s to account for status update delay
     }
   }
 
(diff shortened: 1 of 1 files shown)


app/models/colyseus-models/status.ts
```diff
@@ -238,28 +238,28 @@ export default class Status extends Schema implements IStatus {
...
+      pokemon.status.curseTorment &&
       !pokemon.status.silence
     ) {
       this.triggerSilence(30000, pokemon)
     }
 
-    if (pokemon.effects.has(Effect.CURSE_OF_FATE) && !pokemon.status.curse) {
-      this.triggerCurse(5000)
+    if (pokemon.curseFate && !pokemon.status.curse) {
+      this.triggerCurse(6000)
```
sylvainpolletvillard, line 262: the delay of 5 seconds and not 6 was intentional, to take into account the cooldown of the status update

CostaPrimo: I changed it back to 5s and added a comment saying it is intetional so others doesn't try to correct it like I did.

PR #2026, opened 2024-08-11
Update translations

Ryusei0317:
add synergy_description.


Ryusei0317:
add status_description.

Ryusei0317:
add effect.

mokers-creater (changes requested):
I think this is from SC2 triology, Heart of the swarm. so in my humble opinion, the translations of 'heart of the swarm' should be '군단의 심장', which is Korean translation of SC2 HOTS.

sylvainpolletvillard:
Thanks !

PR #2010, opened 2024-08-08
chore: update translations with Fink

johnkyokim4:

PR #2028, opened 2024-08-11
KR translation : Passive description

mokers-creater:
Passive description has been completed. keep it up, fellows!

-Most of this commits are repeated from #2010 (or should I say #2012) for some reason. I don't know the reason exactly, but I think that's why there are a lot of change logs in this pull-request.
If this causes some crash, please let me know. In that case, I'll clarify my pull-request by removing 'unnecessary' things in my change logs.
Thank you very much.


sylvainpolletvillard:
You're on fire ! Thanks !

PR #2032, opened 2024-08-11
Update translations

Ryusei0317:
add untranslated map and correct mistranslated map.


Ryusei0317:
The commit was not merged well! The part I newly committed was deleted and the existing one was maintained. Can you check it out?

PR #2035, opened 2024-08-11
Update translations(Maps)

Ryusei0317:
Translation about untranslated maps and correction about mistranslated maps.


Ryusei0317:
In request 646ffe2, part of "effect" and "status_description" deletion is not my intend. Is it some kind of error? I didn't delete it from my inlang, but it seems to have been deleted.

mokers-creater:
Here's my hypothesis. 
About 'effect', The parts that are deleted were my work. and you took the part of effect name. I guess when your work about effect, It wasn't successfully merged. Your work was merged successfully, but my previous work about effect wasn't deleted for some reason. so there were two effect_name translation in this file. Deletion of effect_name is simply deletion of doubled strings I guess.

and status_description and synergy_description were just repositioned. They were deleted, but on the other hand, they were remade on the other side of the file.

PR #2071, opened 2024-08-17
Feature: Load more history in user & searched user profile

chenchanglew:
# Summary
- add a new feature load more button to previous history.
- use http get request `/game-history/:playerUid?page=xxx` to load history for player
- update history table in history tsx when receive the request, disable the button if no more history
- add related translation for "load more" in different languages.

image


app/public/src/pages/component/profile/history.tsx
```diff
@@ -43,6 +62,12 @@ export default function History(props: { history: IGameRecord[] }) {
             </div>
           ))}
       </div>
+      <button
+          onClick={loadMore}
+          className="bubbly green create-room-button"
```
sylvainpolletvillard, line 67: wrong classname

app/public/src/pages/component/profile/profile.tsx
```diff
@@ -43,6 +43,31 @@ export default function Profile() {
     dispatch(setSearchedUser(undefined))
     dispatch(setSuggestions([]))
   }, [dispatch])
+  const record_history_num = profile?.history.length;
+
+  const [seed, setSeed] = useState(1);
```
sylvainpolletvillard, line 48: i don't understand that part. seems very convoluted

app/rooms/commands/lobby-commands.ts
```diff
@@ -168,6 +168,85 @@ export class OnJoinCommand extends Command<
...
+        // console.log("load more history for searchedUid: ", searchedUid)
+        const user = await UserMetadata.findOne({ uid: searchedUid })
+        if (user) {
+          const stats = await DetailledStatistic.find(
+            { playerId: user.uid },
+            ["pokemons", "time", "rank", "elo"],
+            { limit: skip+limit, sort: { time: -1 } }
+          )
+          if (stats) {
+            client.send(
```
sylvainpolletvillard, line 188: sending the entire user won't be necessary once we'll have a dedicated http endpoint for game history

app/public/src/pages/component/profile/history.tsx
chenchanglew: ya because this is a new button I created and it doesn't have any css.
should I create a new className for that button with the same css?

app/public/src/pages/component/profile/profile.tsx
chenchanglew: I try to refresh the component when this.state.history is updated.
I couldn't found a way to do that but then I saw this post in stack overflow.
https://stackoverflow.com/questions/56649094/how-to-reload-a-component-part-of-page-in-reactjs

happy to know if there are better solutions to refresh the component (since I am not familiar with react)

sylvainpolletvillard: yeah that answer is really bad, check the second one with useEffect second argument, that's the intended way to do this

app/public/src/pages/component/profile/history.tsx
sylvainpolletvillard: bubbly green should be enough but create-room-button, definitely not appropriate

chenchanglew:
Thanks for the review, the PR should be ready to review again.

PR #2073, opened 2024-08-17
Updated Italian translations

danGuinzo:
Updated the italian translations for some abilities. From Growth to Hidden power !

I noticed 2 possible error for the english translations too, Poison Powder didn't have the special damage type listed in the description, and Dire Claw called for the keyword for the synergy type instead of the poison status.


sylvainpolletvillard:
Correct, thanks !

PR #2076, opened 2024-08-18
RE: #2022 / BUGFIX - Confusion orientation on self target

CostaPrimo:
confusion can no longer target itself as of #2022
The description should match this change.

### DIFF
--- a/app/core/pokemon-state.ts
+++ b/app/core/pokemon-state.ts
@@ -985,15 +985,12 @@ export default class PokemonState {
     return pickRandomIn(candidateCells)
   }
 
-  getTargetCoordinateWhenConfused(
+  getTargetWhenConfused(
     pokemon: PokemonEntity,
     board: Board
-  ): { x: number; y: number } | undefined {
+  ): PokemonEntity {
     let distance = 999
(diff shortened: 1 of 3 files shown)


sylvainpolletvillard:
I wasn't aware of that change, that's a very important nerf for the confusion status and it hasn't been stated in changelog.
IMO should be reverted and fixing orientation problem just with a simple check when coords are the same leading to random orientation.

app/core/pokemon-state.ts
```diff
@@ -1009,20 +1006,15 @@ export default class PokemonState {
...
+          candidatePokemon.push(value as PokemonEntity)
         }
       }
     })
 
-    // Removed as a potential sinner for an orientation error.
-    //candidatesCoordinates.push({ x: pokemon.positionX, y: pokemon.positionY }) // sometimes attack itself when confused
+    candidatePokemon.push(pokemon) // sometimes attack itself when confused
 
-    if (candidatesCoordinates.length > 0) {
```
CostaPrimo, line 1022: List will never be 0 due to self pushed as target, so the check is redundant.

sylvainpolletvillard:
Closing in favor of https://github.com/keldaanCommunity/pokemonAutoChess/pull/2084

PR #2114, opened 2024-08-25
Modified pathfinding algorithm to use A* algorithm, fixing the left-right movement loop bug.

GaryOu101:
This is my attempt at solving pathing issues caused by the current pathfinding algorithm. Since the current one does not account for obstacles(pokemon), the Chebyshev distance scores calculated do not reflect the true optimal pathing. This causes the left-right movement looping seen in the gif below. Moving left is seen as the best move, then right, and so on.

pac1

For my solution, I utilized the A* algorithm which is guaranteed to find the optimal path between two coordinates. During each pokemon movement, I call the findpath() function I created which performs the A* algorithm, returning an array of coordinates of the optimal movements. This function is ran on all outer ring cells that are within pokemon.range's distance to target Pokemon. The shortest path is selected and the first movement in the array is performed. Only 1 movement is used since the optimal path changes with every board change. This process repeats until pokemon is in attacking range of an enemy pokemon.

Below is the same position as the above gif, using the optimized pathfinder. 
pac2

Hopefully this contribution is usable or at least have given some inspiration :)

### DIFF
--- a/app/core/board.ts
+++ b/app/core/board.ts
@@ -164,6 +164,26 @@ export default class Board {
     return cells
   }
 
+  getOuterRangeCells(cellX: number, cellY: number, range = 1, includesCenter = false) {
+    const cells = new Array<Cell>()
+  
+    // Loop through the outer range cells
+    for (let y = cellY - range; y <= cellY + range; y++) {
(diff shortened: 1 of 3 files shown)


zBrodie:
Awesome change!

PR #2117, opened 2024-08-27
Update extremely bad German translation.json

Darkatek7:
I will maybe go on an rewrite a lot of translations. There's no way around that...


sylvainpolletvillard:
Is it clear enough that the ability is hitting one target only ?

Darkatek7:
@sylvainpolletvillard now it's clear nad looks / reads a lot better

PR #2120, opened 2024-08-30
throttle function bug fix and optimization. Disconnect fixes.

ComplexRobot:
Updated the throttle function used when the user presses certain buttons.

This would make it so that after pressing a button, multiple presses would be ignored until after a delay (usually 1 second).

However, it was possible for the function to take longer than a second to execute, meaning it could call the function again before the first one finished leading to possible unintended results and unnecessary processing of additional calls.

So, I added a check that the previous call was not still executing.

Also, the functions passed were often async, so the necessary await code was added to ensure it actually waits for the function to finish before allowing another call.

Hopefully this solves some issues that were happening before and reduces the server load.

A further fix that is needed is to make it a global throttle so that multiple presses of different buttons after the first will be ignored. (My knowledge of typescript is not great, so it might already work that way?)

### DIFF
--- a/app/utils/function.ts
+++ b/app/utils/function.ts
@@ -16,9 +16,9 @@ export function debounce<T extends (...args: any) => any>(
 
 type ThrottledFunction<T extends (...args: any) => any> = (
   ...args: Parameters<T>
-) => ReturnType<T>
+) => Promise<ReturnType<T>>
 
-// fn will be executed if there has not been any previous call during a delay
+// fn will be executed if there has not been any previous call during a delay or if a previous call is not still executing
 export function throttle<T extends (...args: any) => any>(
   fn: T,
   delayInMs: number,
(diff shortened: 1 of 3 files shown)


app/utils/function.ts
```diff
@@ -16,23 +16,28 @@ export function debounce<T extends (...args: any) => any>(
...
-  return function (this: any, ...args: any[]): ReturnType<T> {
+  return async function (this: any, ...args: any[]): Promise<ReturnType<T>> {
     const context = this
 
     if (!inThrottle) {
       inThrottle = true
-      setTimeout(() => (inThrottle = false), delayInMs)
-      lastResult = fn.apply(context, args)
+      // the function could be async and could take longer than the delay to execute
+      let mightBePromise : ReturnType<T> = fn.apply(context, args)
```
sylvainpolletvillard, line 35: i suggest you make another function asyncThrottle since the behavior is quite different

sylvainpolletvillard:
Very interesting additions, thanks !

PR #2123, opened 2024-08-31
Fix Sacred Ash to not resurrect from bench

PAC-Defenestrator:
Fix for this bug: https://discord.com/channels/737230355039387749/1277019643667808266


sylvainpolletvillard:
I know its me that told you about p.isOnBench but that was from my phone and from memory. This function actually doesn't exist, it's actually `isOnBench(p)` imported from pokemon model

Which brings me to the point, please don't make PR if you did not properly tested your changes

PR #2124, opened 2024-08-31
show regional add pick when in appropriate region

PAC-Defenestrator:
Per idea submitted here in discord: https://discord.com/channels/737230355039387749/1278470326732193803

If player is in a region with a regional form of an add pick, the regional form will be proposed instead. The base form is still added to the shop pool.

Tested locally and verified that it works as expected, but let me know if I missed anything or if there's a better way to do any of it.

### DIFF
--- a/app/types/enum/Pokemon.ts
+++ b/app/types/enum/Pokemon.ts
@@ -2472,7 +2472,7 @@ export const PkmFamily: { [key in Pkm]: Pkm } = {
   [Pkm.PAWMOT]: Pkm.PAWMI
 }
 
-export const PkmRegionalVariants: { [key in Pkm]?: Pkm[] } = {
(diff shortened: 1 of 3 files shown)


app/types/enum/Pokemon.ts
```diff
@@ -2484,6 +2484,18 @@ export const PkmRegionalVariants: { [key in Pkm]?: Pkm[] } = {
   [Pkm.SNEASEL]: [Pkm.HISUI_SNEASEL]
 }
 
+export const RegionalPkmBaseForms: { [key in Pkm]?: Pkm } = {
```
sylvainpolletvillard, line 2487: we need to generate this, too easy to forgot to update

sylvainpolletvillard:
Made some changes to simplify code and get rid of manually declared inversed regional variants map

PR #2126, opened 2024-08-31
Change login to prejoin code.

ComplexRobot:
- Changed the "Join Lobby" button so it will now attempt authentication before navigating to the lobby and then reconnect to the room using a reconnection token.
- This will save server resources from needing to load the full lobby page and then boot the user back to the main page after the authentication fails or throws an error.
- I fully tested this to make sure it works as intended.

Edit: Simplified this down to just the main change.

### DIFF
--- a/app/core/pokemon-entity.ts
+++ b/app/core/pokemon-entity.ts
@@ -3,7 +3,7 @@ import { logger } from "../utils/logger"
 import { nanoid } from "nanoid"
 import Count from "../models/colyseus-models/count"
 import Player from "../models/colyseus-models/player"
-import { Pokemon } from "../models/colyseus-models/pokemon"
+import { isOnBench, Pokemon } from "../models/colyseus-models/pokemon"
 import Status from "../models/colyseus-models/status"
 import PokemonFactory from "../models/pokemon-factory"
 import { getSellPrice } from "../models/shop"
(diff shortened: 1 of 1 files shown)


sylvainpolletvillard:
I'm not sure if joining lobby room twice is actually saving resources. What heavy requests do we do on lobby page before room is joined ?

Also if pre-check lobby join is accepting me, what gives me the guarantee the second lobby join on lobby page will be accepted and not rejected due to max concurrent users limit ? Currently many users are spamming the join button all day, so there's a high chance the ccu count changes in between the two joinLobby calls

sylvainpolletvillard:
Also the default value for leave() consented is true so these changes are unnecessary
https://docs.colyseus.io/client/?h=leave#leave-consented-boolean
image

app/public/src/pages/preparation.tsx
```diff
@@ -212,16 +218,19 @@ export default function Preparation() {
           const game: Room<GameState> = await client.joinById(roomId, {
             idToken: token
           })
+          logger.log(game)
```
sylvainpolletvillard, line 221: to be removed

sylvainpolletvillard:
> The basic layout of the page is loaded before being authenticated into the room. It may not seem like much, but thousands or hundreds of these being downloaded from the server adds up.

This is client-side rendering, there's no additional server requests apart the avatars that come from raw.githubusercontent.com and a few other assets that are cached in Service Worker. Test it by yourself, sniff the network.

> In other words, it still only does a single authentication and waits for the user to reconnect (does not actually remove them from the lobby while it's waiting - I've set the timeout to 30 seconds, but that can be changed). That's entirely the point of what I did! It will skip authentication if it authenticated you on the main page.

Fair enough, this 30 seconds timeout could actually help a bit with the "join button spamming" situation we have currently, in addition to the throttle.

PR #2127, opened 2024-08-31
add close codes and error messages

sylvainpolletvillard:
for all forced room disconnections, add close codes with proper translated error messages

added clear error messages for several cases of disconnections, including:
- inactivity on lobby
- room deleted
- room already full
- room with not enough players
- user authentication expired
- kick
- ban

### DIFF
--- a/app/rooms/commands/preparation-commands.ts
+++ b/app/rooms/commands/preparation-commands.ts
@@ -20,6 +20,7 @@ import { cleanProfanity } from "../../utils/profanity-filter"
 import { pickRandomIn } from "../../utils/random"
 import { entries, values } from "../../utils/schemas"
 import PreparationRoom from "../preparation-room"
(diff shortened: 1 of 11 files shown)


app/rooms/commands/preparation-commands.ts
```diff
@@ -438,12 +434,7 @@ export class OnDeleteRoomCommand extends Command<
       const user = this.state.users.get(client.auth?.uid)
       if (user && [Role.ADMIN, Role.MODERATOR].includes(user.role)) {
         this.room.clients.forEach((cli) => {
-          cli.send(Transfer.KICK)
-          cli.leave()
-        })
-        this.room.clients.forEach((cli) => {
```
sylvainpolletvillard, line 444: i'm assuming this was a mistake to kick all clients twice

PR #2137, opened 2024-09-02
Change prep room reconnect wait time to 30 seconds.

ComplexRobot:
Changing the wait time of the `onLeave` function to match the expiry time of the reconnection token.
Hopefully this will allow more time for slow reconnects and less connection drops when joining a pre-game room.
(The previous time was 3 seconds, which seems to fail often.)

Small side note: these type of reconnect tokens don't appear to need to have an expiry time set on them, because the timeout is handled by the `onLeave` function.
On second thought, maybe it is better to have those expire to not keep stale tokens around.


sylvainpolletvillard:
>  these type of reconnect tokens don't appear to need to have an expiry time set on them
an expiration time is needed client-side to avoid showing the reconnection modal if the token is already expired for a long time ago

> I think also, there needs to be a timeout check that periodically checks if the connection was dropped.
Websocket is supposed to deal with that and maintain its connection. If dropped completely, we go through the onLeave callback and can detect it from there. No need for polling/heartbeat with websocket

sylvainpolletvillard:
I think 30 seconds is too much to block a seat in a lobby. These days, a lobby can be filled in less than 10 seconds.

But 3 seconds is probably too short with the current affluence and lag. How about 10 seconds ?

sylvainpolletvillard:
> In that case, how do we detect when the connection just ends unexpectedly and alert the user or reconnect automatically?

I told you, with the onLeave callback. It receives the websocket closing code as argument

PR #2139, opened 2024-09-02
Quick play tries to reconnect to game before finding a new room.

ComplexRobot:
Will try to reconnect to a quick play room that the user is already in before searching for a new room in case of an error that canceled the first join.

### DIFF
--- a/app/public/src/pages/component/available-room-menu/available-room-menu.tsx
+++ b/app/public/src/pages/component/available-room-menu/available-room-menu.tsx
@@ -125,6 +125,29 @@ export default function AvailableRoomMenu() {
   }, 1000)
 
   const quickPlay = throttle(async function quickPlay() {
+    const reconnectToken: string = localStore.get(LocalStoreKeys.RECONNECTION_TOKEN)
+    if (reconnectToken) {
+      try {
+        const room: Room<PreparationState> = await client.reconnect(reconnectToken)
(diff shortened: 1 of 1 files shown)


sylvainpolletvillard:
Doesn't this introduce a side-effect to show the person to other lobby members a fraction of second joining/leaving then rejoining again ? With all the message/sound effects spam associated ? And again because of the reconnection delay, it is possible that the lobby is full in the meantime between the two joins

sylvainpolletvillard:
I think if a reconnection token to a preparation room is detected from lobby page, we should use the reconnection proposal modal to let the user reconnect if he wants. If he clicks no, the reconnection token is deleted and quick play button will have the normal behavior.

sylvainpolletvillard:
> No, the reconnection will only work if they are already in the room, not disconnected. This is a rare edge case that could happen. If they are on the lobby page but due to lag or something, they are technically connected to a quick play room.
That way, they will join the quick play room they are already in rather than a random one if they press the quick play button.
It's also the only way they will join a full 8/8 players quick play room that they are in with the current code.

OK I understand now, yes this is very much needed to avoid the "ghost game" issue many players complain about

sylvainpolletvillard:
Yes with a 10 seconds reconnect window, you can assume it's only happening when a player is spamming the quick play button and leave button. In that case, reconnecting to the same lobby without asking seems appropriate 👍

PR #2140, opened 2024-09-02
Cleanup stale reconnect token on lobby reconnect.

ComplexRobot:
Deleting the invalid reconnect token after reconnecting to prevent reconnection errors to other rooms.


sylvainpolletvillard:
>  It would be better to split them up by room type. i.e., LocalStoreKeys.RECONNECTION_TOKEN should be split into multiple room types.
Agree. You can do it if you want

> This is now forcing the page to reload twice. useEffect already forced a reload once, and now useState reloads it again. So, the reconnect will try to reconnect when it's already connected causing the "seat reservation" error.

By page reload i suppose you mean rerender. Components render a lot with React, it's not a big deal assuming effects are manage d correctly

PR #2141, opened 2024-09-02
Change lobbyJoined back to useRef

ComplexRobot:
Changes added in #2132 are causing an error with the login reconnect.
Specifically these changes:
```ts
// app\public\src\pages\lobby.tsx
const [lobbyJoined, setLobbyJoined] = useState<boolean>(false)
//...
if (!lobbyJoined) {
//...
setLobbyJoined(true)
```
This is now forcing the page to reload twice. `useEffect` already forced a reload once, and now `useState` reloads it again. So, the reconnect will try to reconnect when it's already connected causing the "seat reservation" error.
`useRef` was fine, I think. It's the same function as `useState` except it doesn't force the page to reload when the variable changes.

Any variable referenced in `useEffect` will force a refresh when it changes.
```ts
// variable being changed
lobbyJoined.current = true
// ...
// variables that are referenced
}, [lobbyJoined, dispatch])
```

### DIFF
--- a/app/public/src/pages/lobby.tsx
+++ b/app/public/src/pages/lobby.tsx
@@ -242,22 +242,23 @@ export async function joinLobbyRoom(
         dispatch(logIn(user))
         try {
           const token = await user.getIdToken()
-
-          const reconnectToken: string = localStore.get(LocalStoreKeys.RECONNECTION_TOKEN)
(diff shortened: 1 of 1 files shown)


sylvainpolletvillard:
I moved back to useRef in master for lobbyJoined

PR #2148, opened 2024-09-03
Fix prejoin lobby code. Separate lobby token. Allow reconnection when leaving to other room.

ComplexRobot:
Fixed and tested the prejoin lobby code with strict mode off.
- ~~Send the user profile on reconnect.~~
- Save reconnection key under a separate name specific to the lobby, `LocalStoreKeys.RECONNECTION_LOBBY`.
- Save the reconnection token for 5 minutes after successfully joining the lobby.
- Consent false when leaving lobby to give 30 seconds to rejoin without having to re-authorize.


app/public/src/pages/lobby.tsx
```diff
@@ -242,25 +242,25 @@ export async function joinLobbyRoom(
...
-              logger.log(error)
-              localStore.delete(LocalStoreKeys.RECONNECTION_TOKEN)
+              localStore.delete(LocalStoreKeys.RECONNECTION_LOBBY)
             }
           }
-          const lobby = store.getState().network.lobby
           const room: Room<ICustomLobbyState> = lobby ?? await client.join("lobby", {
             idToken: token
           })
+          // store key for 5 minutes, but can still only rejoin within 30 seconds of leaving
```
sylvainpolletvillard, line 262: can you explaining the reasoning about the 5 minutes client side token expiration ?

ComplexRobot: It's kind of an arbitrary number, but I think it should be at least 1 minute.

Basically when the user is idling in the lobby, looking through their profile/booster packs/etc. the local token is still counting down. Or if they join a preparation room and get kicked/lag out or otherwise decide to leave back to the lobby, during that time the token would be counting down.

The token could expire even though they are still within the 30 second rejoin window.
This gives a little buffer of time for them to still be able to rejoin.
Logically, the local token could never expire and everything would work fine, but I feel there should be a cutoff for weird edge cases.

ComplexRobot: Otherwise, if it's a small number you might be forcing more login/logout requests and poor user experience I feel.

Also, to be clear, an idle kick is a separate issue that this doesn't address.

sylvainpolletvillard: Currently we have a cron job that runs every minute and kick all players that habe been in lobby room for more than one minute. So this expiration time doesn't seem very useful at the moment. It's not really a problem, just I could see it much longer, anyway server will be authrotitative when it comes to kicking inactive players

ComplexRobot: Then, there's no point for it to be longer than 1 minute, I suppose.

sylvainpolletvillard: 2 minutes max, if you're desynced with the cron job tick. But as I said, doesn't change much, it could be one hour or one day

app/public/src/pages/lobby.tsx
```diff
@@ -240,25 +240,26 @@ export async function joinLobbyRoom(
...
         try {
           const token = await user.getIdToken()
-
-          let lobby: Room<ICustomLobbyState> | undefined
-          const reconnectToken: string = localStore.get(LocalStoreKeys.RECONNECTION_TOKEN)
+          let lobby: Room<ICustomLobbyState> | undefined = store.getState().network.lobby
+          if (lobby?.connection.isOpen) {
+            localStore.set(LocalStoreKeys.RECONNECTION_LOBBY, lobby.reconnectionToken, 30)
+            lobby.onLeave.clear()
+            await lobby.leave(false)
```
sylvainpolletvillard, line 247: What's the point of leaving the lobby room if you reconnect immediately to it a few lines after ?

sylvainpolletvillard: Couldn't this be simplified to:
```
 firebase.auth().onAuthStateChanged(async (user) => {
      if (user) {
        dispatch(logIn(user))
        try {
          const token = await user.getIdToken()
          let room: Room<ICustomLobbyState> | undefined = store.getState().network.lobby
          const reconnectToken: string = localStore.get(LocalStoreKeys.RECONNECTION_LOBBY)
          if ((!room || !room?.connection.isOpen) && reconnectToken) {
            try {
              room = await client.reconnect(reconnectToken)
            } catch (error) {
              localStore.delete(LocalStoreKeys.RECONNECTION_LOBBY)
            }
          }
          if(!room) {
            room = await client.join("lobby", {
              idToken: token
            })
            // store key for 5 minutes, but can still only rejoin within 30 seconds of leaving
            localStore.set(LocalStoreKeys.RECONNECTION_LOBBY, room.reconnectionToken, 60 * 5)
          }
```

ComplexRobot: This is because the user profile must be set.
It can only be set from a `client.send` message from the server either in `onAuth` or `onLeave`.
It's just forcing that message to be sent.

ComplexRobot: So, either `reconnect` or `join` must be called no matter what.

app/rooms/custom-lobby-room.ts
```diff
@@ -459,11 +459,8 @@ export default class CustomLobbyRoom extends Room<LobbyState> {
       if (consented) {
         throw new Error("consented leave")
       }
-      const userProfile = this.users.get(client.auth.uid)
-      if (!userProfile) {
-        throw new Error("Missing User Profile.")
-      }
       await this.allowReconnection(client, 30)
+      const userProfile = this.users.get(client.auth.uid)
       client.send(Transfer.USER_PROFILE, userProfile)
```
ComplexRobot, line 464: Here is where the `reconnect` is handled and the user profile is sent.

ComplexRobot: It's set up at line 420 in the `onAuth` method where `join` is handled.

app/public/src/pages/lobby.tsx
ComplexRobot: That's why I've also set lobby to `undefined` after that, so if the `reconnect` fails, `join` will for sure be called.

sylvainpolletvillard: then what's the point to use a reconnection token in the first place if we need to force leaving and rejoining

sylvainpolletvillard: If i count correctly, with the prejoin from auth page, that's joining/leaving the lobby room 3 times in a row just for one login

ComplexRobot: Well, this particular check only happens if the user profile fails to be set. So only when a fresh new account logs in for the first time.
But, it's not technically leaving the lobby. It's `leave(false)` which is just to put it on standby for the `reconnect` function.
The way it works is just with a reconnection token as far as I know.
The point is `reconnect` is less overhead than `join` and should be preferred to be used.

ComplexRobot: > If i count correctly, with the prejoin from auth page, that's joining/leaving the lobby room 3 times in a row just for one login

I know the function is called `leave` but it doesn't actually leave. It just does a particular algorithm.
In this case, we're using it to transfer state between pages and access a user profile message from the server.

ComplexRobot: Also, using `reconnect` we maintain a connection instead of constantly freeing and allocating a new connection and constant logging in and out of users and all the overhead of allocating a new user.

ComplexRobot: Whenever a user joins it will have to allocate the data of that connection, initialize it all, do whatever setup code is needed. Then, then they disconnect, it will have to do any cleanup code and set all the data to be freed by the garbage collector.

If you're constantly doing this, the garbage collector is going to be overtaxed and you're going to see what looks like a memory leak, but the garbage collector is just going to get stuck with way too many tasks.

ComplexRobot: If you use reconnect, you're avoiding those two steps and doing close to nothing.

ComplexRobot: All of the overhead will basically be handled client-side.

sylvainpolletvillard:
Honestly the authentication code became **so much more complicated** since this prejoin thing was added, I can no longer comprehend how this whole process works

I just notice when testing locally the PR branch that i got constantly "Error: seat reservation expired." both in server console and browser console.

This feels wrong to me, but maybe that's beyond my understanding. In any case it probably lacks comments to remain maintainable on the long term

sylvainpolletvillard:
> Why start now?

Because I understand all the rest of the code but not yours. Even with your explainations it's still not very clear to me

PR #2150, opened 2024-09-03
Separate reconnection tokens by room type. Save reconnection token and room ID.

ComplexRobot:
Fully separated the `RECONNECTION_TOKEN` into a separate token for each room type.
Also changed setting the value to include the room ID for easy lookup.


sylvainpolletvillard:
>  I noticed there's a bug with the reconnect to game modal. It doesn't properly connect to the game, saying it doesn't exist or it will navigate back to the main page and the music for the game starts playing.
this is a regression, used to work just fine

PR #2160, opened 2024-09-04
Change `connection.close` to `await leave(false)` (urgent)

ComplexRobot:
Fixes a huge bug with joining rooms that results in a race condition causing the "seat reservation expired" error.

**Explanation**

`Room.reconnect` waits for an `allowReconnection` call in `onLeave` to successfully reconnect.
Example:
```ts
async onLeave(client: Client, consented: boolean) {
  try {
    // if true or no parameter is passed to a leave function, leave immediately
    if (consented) {
      throw new Error("consented leave")
    }
    // if false was passed to leave, wait for a reconnection initiated by client.reconnect
    await this.allowReconnection(client, 30)
  } catch (error) {
    // if the reconnect failed, an error is caught and OnLeaveCommand processed
    this.dispatcher.dispatch(new OnLeaveCommand(), { client })
  }
}
```

- `leave()` starts the leave process and calls `onLeave`.
- `connection.close()` is equivalent to calling `leave(false)` and allows for a reconnection.
- `leave()` is an async function that waits for the `onLeave` process to begin before yielding.
- `connection.close()` is not async and returns without waiting.

Therefore, this caused a race condition in cases like this:
```ts
// join a room
const room = await client.join("preparation", { idToken })
// save a reconnection token to reconnect later
localStore.set(
  LocalStoreKeys.RECONNECTION_PREPARATION,
  { reconnectionToken: room.reconnectionToken, roomId: room.roomId },
  30
)
// check if connection is open and close it to wait for reconnect
if (room.connection.isOpen) {
  room.connection.close() // should be -> await room.leave(false)
}
// send user to the preparation page to reconnect
navigate("/preparation")
//...

// in the preparation page:

// !!! this code could be reached BEFORE onLeave is called <- RACE CONDITION

// get the previously saved reconnection token
const cachedReconnectionToken = localStore.get(LocalStoreKeys.RECONNECTION_PREPARATION)?.reconnectionToken
// attempt to reconnect to the room using the saved token - allowing allowReconnection to return without error
const room = await client.reconnect(cachedReconnectionToken)
```

Changing the `connection.close()` to `await leave(false)` eliminates the race condition by ensuring it waits for `onLeave` to start before moving on.

In general, it is probably better practice to use `leave` instead of `connection.close` to mark your intent. (Whether you meant to close the connection immediately or are planning to reconnect after.)

The equivalent of `connection.close()` is `leave(false)` without an `await`, therefore `connection.close` can always be replaced by a call to `leave` instead.

I wrote this detailed explanation so there's hopefully no confusion as to why I'm making this change.


sylvainpolletvillard:
If you go with `room.leave()` you don't need all the `if connection open` checks, it's built in the method: https://github.com/colyseus/colyseus.js/blob/master/src/Room.ts#L98

sylvainpolletvillard:
OK i don't think this error message is a big deal but checking isOpen is fine then. LGTM !

PR #2165, opened 2024-09-05
Automatic reconnect when lag or timeout in prep room and in game

ComplexRobot:
Added a fix which auto reconnects when the connection ends unexpectedly or times out.
- No more preparation room getting stuck with no indication that you lagged out.
- No more game loading screen hanging until you manually reload it.
- No more game freezes and hangs until you manually reload it.

This now just for the game. Preparation room was split into another PR.

### DIFF
--- a/app/public/src/pages/game.tsx
+++ b/app/public/src/pages/game.tsx
@@ -109,7 +109,7 @@ export default function Game() {
   const [finalRankVisible, setFinalRankVisible] = useState<boolean>(false)
   const container = useRef<HTMLDivElement>(null)
 
-  const MAX_ATTEMPS_RECONNECT = 10
+  const MAX_ATTEMPS_RECONNECT = 3
 
   const connectToGame = useCallback(
     async (attempts = 1) => {
--- a/app/public/src/pages/game.tsx
+++ b/app/public/src/pages/game.tsx
@@ -25,6 +25,7 @@ import { DungeonDetails } from "../../../types/enum/Dungeon"
 import { Team } from "../../../types/enum/Game"
 import { Pkm } from "../../../types/enum/Pokemon"
 import { getFreeSpaceOnBench } from "../../../utils/board"
+import { CloseCodes, CloseCodesMessages } from "../../../types/enum/CloseCodes"
 import { logger } from "../../../utils/logger"
 import { addWanderingPokemon } from "../game/components/pokemon"
 import GameContainer from "../game/game-container"
@@ -108,7 +109,7 @@ export default function Game() {
   const [finalRankVisible, setFinalRankVisible] = useState<boolean>(false)
   const container = useRef<HTMLDivElement>(null)
 
-  const MAX_ATTEMPS_RECONNECT = 3
+  const MAX_ATTEMPS_RECONNECT = 10
 
   const connectToGame = useCallback(
     async (attempts = 1) => {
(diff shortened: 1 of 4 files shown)


app/public/src/pages/game.tsx
```diff
@@ -292,6 +294,24 @@ export default function Game() {
...
       }) as EventListener)
 
+      room.onLeave((code) => {
+        const shouldReconnect = code === CloseCodes.ABNORMAL_CLOSURE || code === CloseCodes.TIMEOUT
+        logger.info(`left game room with code ${code}`, { shouldReconnect })
+        if (shouldReconnect) {
+          logger.log("Connection closed unexpectedly or timed out. Attempting reconnect.")
+          // Clear out all game data to reload the game
+          connected.current = false
+          initialized.current = false
```
sylvainpolletvillard, line 304: I'm pretty sure this registers several times the same event listeners on gameElm

app/public/src/pages/game.tsx
```diff
@@ -292,6 +294,24 @@ export default function Game() {
...
+          if (gameContainer && gameContainer.game) {
+            gameContainer.game.destroy(true)
```
sylvainpolletvillard, line 306: i don't see why a socket disconnection should imply destroying and recreating the game canvas. 

The only thing needed to reconnect after a websocket abnormal closure is `connectToGame` and set up the event listeners on the room. Therefore all the `room.onMessage` and `room.state.listen` could be moved to `connectToGame` after room is reconnected, so that the only thing you need to do is `connected.current = false`

app/public/src/pages/game.tsx
```diff
@@ -292,6 +294,24 @@ export default function Game() {
...
+          }
+          setLoaded(false)
+          dispatch(leaveGame())
+          // Update dependency variable to force a re-render
+          setReconnectTries(x => x + 1)
```
sylvainpolletvillard, line 311: you don't need to re-render. There's nothing to update in the component render, it's just reconnecting to room

app/public/src/pages/preparation.tsx
```diff
@@ -201,6 +204,18 @@ export default function Preparation() {
...
+          // Restart the expiry timer of the reconnection token and clear room data to refresh
+          localStore.set(
+            LocalStoreKeys.RECONNECTION_PREPARATION,
+            { reconnectionToken: r.reconnectionToken, roomId: r.roomId },
+            30
+          )
+          initialized.current = false
+          dispatch(leavePreparation())
+          // Update dependency variable to force a re-render
+          setReconnectTries(x => x + 1)
```
sylvainpolletvillard, line 218: same comment than for game, you don't need to rerender. It's much easier for preparation, you can just call the `reconnect()` function that takes care of everything. We could do a similar `reconnect` function for game room as well

app/public/src/pages/game.tsx
ComplexRobot: I agree. I had some errors when I tried that, though.

ComplexRobot: > Therefore all the `room.onMessage` and `room.state.listen` could be moved to `connectToGame` after room is reconnected, so that the only thing you need to do is call `connectToGame()` again

Oh! Good idea.

app/public/src/pages/game.tsx
ComplexRobot: I had some error with phaser I didn't know how to solve.

app/public/src/pages/preparation.tsx
ComplexRobot: Right. I think what it is, is it needs to reconnect, set up all the event listeners again and set the state variable to the new room object.

sylvainpolletvillard:
OK for prep room, provided you remove your setTimeout testing code

app/public/src/pages/game.tsx
```diff
@@ -292,379 +297,397 @@ export default function Game() {
...
+    room.onLeave((code) => {
+      const shouldReconnect = code === CloseCodes.ABNORMAL_CLOSURE || code === CloseCodes.TIMEOUT
+      logger.info(`left game room with code ${code}`, { shouldReconnect })
+      if (shouldReconnect) {
+        logger.log("Connection closed unexpectedly or timed out. Attempting reconnect.")
+        connected.current = false
+        dispatch(leaveGame())
+        connectToGame()
+      }
+    })
```
ComplexRobot, line 328: Everything below here is the same as before. (Diff seems to be confused by the unindent.

PR #2166, opened 2024-09-05
Asynchronize all double leave calls

ComplexRobot:
A simple optimization, make all of the double leave calls execute simultaneously instead of one after the other using `Promise.allSettled`.

### DIFF
--- a/app/public/src/pages/game.tsx
+++ b/app/public/src/pages/game.tsx
@@ -221,11 +221,8 @@ export default function Game() {
     }
     dispatch(leaveGame())
     navigate("/after")
-
-    try {
-      await room?.leave()
-    } catch (error) {
-      logger.warn("Room already closed")
+    if (room?.connection.isOpen) {
+      room.leave()
     }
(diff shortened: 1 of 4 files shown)


app/public/src/pages/game.tsx
```diff
@@ -221,11 +221,8 @@ export default function Game() {
     }
     dispatch(leaveGame())
     navigate("/after")
-
-    try {
-      await room?.leave()
-    } catch (error) {
-      logger.warn("Room already closed")
+    if (room?.connection.isOpen) {
```
ComplexRobot, line 224: A simple optimization, return immediately instead of `await`, since nothing happens after.

app/public/src/pages/game.tsx
```diff
@@ -221,11 +221,8 @@ export default function Game() {
...
+      room.leave()
```
ComplexRobot, line 225: A simple optimization, return immediately instead of `await`, since nothing happens after.

PR #2167, opened 2024-09-05
move cron jobs from pokemonAutoChessEloDecay

sylvainpolletvillard:
move jobs from https://github.com/keldaanCommunity/pokemonAutoChessEloDecay/ inside the project and run them with cron time

Paris time, every day:
8AM : delete old anonymous accounts
8:05AM: delete old history
8:10 AM; apply elo decay
8:15 AM; recompute title statistics

Many parameters from these jobs have been externalized as constants in Config

This makes it much more easier to apply these tasks automatically for community servers instead of github actions


app/index.ts
keldaan-ag, line 1: Why about creating a Cron job for the fetch leaderboard

sylvainpolletvillard: It was a cron job before. I replaced it by a setInterval because fetching leaderboard is required for all node instances, and not just once per server like all other cron jobs.

PR #2185, opened 2024-09-09
Add automatic reconnect logic to game

ComplexRobot:
The simplest code needed for an automatic reconnect.

Directly reconnects the web socket without creating a new room instance or affecting any other game system.

### DIFF
--- a/app/public/src/game/reconnect-logic.ts
+++ b/app/public/src/game/reconnect-logic.ts
@@ -68,7 +68,7 @@ export function enableAutoReconnect<T = any>(client: Client, room: Room<T>, reco
   }
 
   // update to new reconnection token after reconnecting
-  room.onJoin(() => {
+  (room as any).onJoin(() => {
     localStore.set(
       reconnectStoreKey,
       { reconnectionToken: room.reconnectionToken, roomId: room.roomId },
--- a/app/public/src/game/reconnect-logic.ts
+++ b/app/public/src/game/reconnect-logic.ts
@@ -72,7 +72,7 @@ export function enableAutoReconnect<T = any>(client: Client, room: Room<T>, reco
   transport.ws.onmessage = room.connection.events.onmessage = function onmessage(messageEvent: MessageEvent) {
     originalOnmessage(messageEvent)
     const code = Array.from(new Uint8Array(messageEvent.data))[0]
-    if (code == Protocol.JOIN_ROOM) {
+    if (code === Protocol.JOIN_ROOM) {
       localStore.set(
         reconnectStoreKey,
         { reconnectionToken: room.reconnectionToken, roomId: room.roomId },
--- a/app/public/src/game/reconnect-logic.ts
+++ b/app/public/src/game/reconnect-logic.ts
@@ -1,0 +1,88 @@
+import { Client, Room } from "colyseus.js"
+import { WebSocketTransport } from "colyseus.js/lib/transport/WebSocketTransport"
+import { logger } from "../../../utils/logger"
+import { CloseCodes } from "../../../types/enum/CloseCodes"
+import { Transfer } from "../../../types"
+import { localStore, LocalStoreKeys } from "../pages/utils/store"
+
+const MAX_RECONNECT_TRIES = 3
+
+/**
+ * Automatically reconnects when the connection closes unexpectedly or times out.
+ * @param client The colyseus.js client-side Client.
+ * @param room The colyseus.js client-side Room.
+ * @param reconnectStoreKey The localStorage key string to save the reconnection token.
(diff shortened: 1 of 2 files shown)


app/public/src/game/reconnect-logic.ts
```diff
@@ -0,0 +1,93 @@
...
+        reconnectStoreKey,
+        { reconnectionToken: room.reconnectionToken, roomId: room.roomId },
+        tokenExpirationTime
+      )
+    }
+  }
+
+  // wrap send method with a connection check
+  const originalSend = room.send
+  room.send = function send<T = any>(type: string | number, message?: T): void {
```
sylvainpolletvillard, line 86: overloading colyseus methods is a bad practice

app/public/src/game/reconnect-logic.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    logger.log("Connection closed unexpectedly or timed out. Attempting to reconnect.")
+
+    const token = room.reconnectionToken.slice(room.reconnectionToken.indexOf(":") + 1)
+
+    const reconnect = function reconnect(tries = 1) {
+      logger.log(`Reconnect try ${tries}/${MAX_RECONNECT_TRIES}`)
+
+      // request reconnect from the server
+      client.http.post(`matchmake/reconnect/${room.roomId}`, {
```
sylvainpolletvillard, line 36: is this public API ?

app/public/src/game/reconnect-logic.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @param room The colyseus.js client-side Room.
+ * @param reconnectStoreKey The localStorage key string to save the reconnection token.
+ * @param tokenExpirationTime The expiration time in seconds for the localStorage key expiration.
+ */
+export function enableAutoReconnect<T = any>(client: Client, room: Room<T>, reconnectStoreKey: LocalStoreKeys, tokenExpirationTime: number) {
+
+  // override web socket onclose event with a custom function
+  const transport: WebSocketTransport = room.connection.transport as any
+  const originalOnclose: (event: CloseEvent) => any = transport.ws.onclose as any
+  transport.ws.onclose = room.connection.events.onclose = function onclose(closeEvent: CloseEvent) {
```
sylvainpolletvillard, line 22: overloading colyseus methods is a bad practice

app/public/src/game/reconnect-logic.ts
ComplexRobot: I did this, because there is no other way.

app/public/src/game/reconnect-logic.ts
ComplexRobot: It is public, and is part of the API, so, technically, yes.

app/public/src/game/reconnect-logic.ts
ComplexRobot: Okay, this one. You could technically add an if check like 100 times to the code. I figured this was easier to maintain.

app/public/src/game/reconnect-logic.ts
ComplexRobot: This one I did because there's no other way to do this.

app/public/src/game/reconnect-logic.ts
sylvainpolletvillard: can you point to documentation ? can't find anything

app/public/src/game/reconnect-logic.ts
```diff
@@ -0,0 +1,93 @@
...
+      // receive server response
+      }).then((response) => {
+        const responseData = response.data
+        if (responseData.error) {
+          throw new Error(responseData.error)
+        }
+        // create url endpoint and connect web socket
+        const roomData = responseData.room
+        const sessionId = responseData.sessionId
+        const endpoint = `${window.location.protocol.replace("http", "ws")}//${window.location.hostname}${(window.location.port && `:${window.location.port}`)}`
```
sylvainpolletvillard, line 52: this path construction is very likely to break

app/public/src/game/reconnect-logic.ts
ComplexRobot: I found this in the source code. Same as a lot of undocumented things like the `connection` property.

sylvainpolletvillard:
It looks like a patch above colyseus. Could be something to discuss on Colyseus discord, I'm sure this problem has been tackled by other people. Overriding colyseus code and using semi-public API is a straight no for me though.

app/public/src/game/reconnect-logic.ts
sylvainpolletvillard: so it's not public...

sylvainpolletvillard:
I don't want it to break as soon as i update colyseus

PR #2201, opened 2024-09-12
Add Thai index

Zeleos753:


sylvainpolletvillard:
Is thai language ready enough for that ?

Zeleos753:
> Is thai language ready enough for that ?

I think it is. I already make an announcement in my server (Thai Community) that I'm not going to translate (below)
and no one against it many player say only focus for game aspect like skill, synergy, passive description

Pokemon name > since some game/anime use Japanese name some use English name
Skill name, Item > game/tcg/anime some use Thai some use English some use Japanese
Map > because it's really weird if translate in Thai many game/anime always just use default name
Role > we use term admin, mod, staff from English word anyway

so I going to let all other thing that not translated in English for not confusing player

sylvainpolletvillard:
Alright !

PR #2203, opened 2024-09-12
Update translations

cesuura:
I'm working on matching the abilities with the same names than the original moves


sylvainpolletvillard:
Thanks !

PR #2207, opened 2024-09-13
cache busting for translations files

sylvainpolletvillard:
Try to fix the cache issue with translations file on patch release. No more Ctrl+F5 !

At first i tried to do filename hash suffix with esbuild like what we do for index.js and index.css, but it ends up being quite complicated. esbuild is not really the tool for that.

I made hash generation based on file content with one hash specific to each translation file, but then the problem was to decide where to store those hashes in order for i18next to load them properly. Storing these hashes in a global variable or concatenating a variable declaration in the built index.js file was pretty ugly to me, so instead i decided to use the hash from index.js to suffix all translations. My reasoning is that the main index.js file will be very likely updated every time  we ship a hotfix and need people to reload translations. I know its not perfect, for example we'll still have the cache if we deliver exclusively a translation hotfix without touching at the source.

Another option is to rely on service worker exclusively to cache these translation files, and remove all cache headers from server for those files. It's much simpler but also force to change version of service worker every time we touch at translations, redownloading all assets.


keldaan-ag:
we can give this a try

keldaan-ag:
👍

sylvainpolletvillard:
closing in favor of https://github.com/keldaanCommunity/pokemonAutoChess/pull/2235

PR #2220, opened 2024-09-18
Fix ENTRAINMENT doens't effect properly

Zeleos753:
discord: https://discord.com/channels/737230355039387749/1285705073367715891

• potential target check target team instead
• check if it does not have Entrainment ability
• sort and execute if list not empty
• sort range from target instead from user*

*word state pick _another enemy nearby_ i assume it nearby target instead of user


sylvainpolletvillard:
Thank you !

PR #2221, opened 2024-09-18
Fix Duke Title given on wrong reason

Zeleos753:
discord: https://discord.com/channels/737230355039387749/1285908004843290624
from check how many data in player pokemonCollection size since player getting add $index after got shard. that make it if player have 30 different types of shards instead of profiles

change to count how profile player actually has


Zeleos753:
spacing is really weird i think i t

PR #2236, opened 2024-09-23
update regional pool on add pick

sylvainpolletvillard:
fix https://discord.com/channels/737230355039387749/1287696664227745847

### DIFF
--- a/app/rooms/game-room.ts
+++ b/app/rooms/game-room.ts
@@ -1008,6 +1008,9 @@ export default class GameRoom extends Room<GameState> {
         this.state.shop.addAdditionalPokemon(pkm)
       }
 
+      // update regional pokemons in case some regional variants of add picks are now available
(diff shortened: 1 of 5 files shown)


app/rooms/game-room.ts
```diff
@@ -1008,6 +1008,9 @@ export default class GameRoom extends Room<GameState> {
         this.state.shop.addAdditionalPokemon(pkm)
       }
 
+      // update regional pokemons in case some regional variants of add picks are now available
+      this.state.players.forEach((p) => p.updateRegionalPool(this.state, false))
```
sylvainpolletvillard, line 1012: i hesitate to update all player regional pokemons whenever an add pick is picked, or only at the end of the additional pick picking phase ; just to avoid updating 8 times in a row regional pokemons in a matter of seconds

PR #2240, opened 2024-09-25
fix sudowoodo line gaining atk bug

Zeleos753:
https://discord.com/channels/737230355039387749/1281924087295512606

after last fix player report and I already tested that can be happening if pokemon get status freeze/sleep

since status.tree pokemon will set to Idle state and if pokemon got **freeze** also send to idle state in **update** it getting loop gaining damage than it should be

the thing is I'm not completely sure will affect  another situation so i not PR this. but after test awhile i think it should be ok


sylvainpolletvillard:
I don't understand how that change is fixing the bug, it just sets the action to Sleep instead of Idle ?

PR #2256, opened 2024-09-28
Fix spike armor (again)

sylvainpolletvillard:
Another fix found for spike armor: it was not dealing damage to ranged pokemons at melee contact


keldaan-ag:
👍

PR #2266, opened 2024-09-30
Fix Evolution on Mothim

HametAk:
https://discord.com/channels/737230355039387749/1290341692830121984/1290341692830121984

When evolving Wormadam, we need to check the last pokemon used for the evolution, since it's possible to have every single Wormadam in your team.

So when we have WTrash, WPlant and WGround we check the last used pokemon for the evolution and add the correct Type to Mothim.

Furthermore we also change the order of the evolution rule, so we acquire pokemon first and then calculate the synergies

### DIFF
--- a/app/models/colyseus-models/pokemon.ts
+++ b/app/models/colyseus-models/pokemon.ts
@@ -13747,6 +13747,27 @@ export class WormadamPlant extends Pokemon {
     const regionSynergies = DungeonDetails[map]?.synergies
     return regionSynergies.includes(Synergy.GRASS)
   }
+
+  onEvolve({
+    pokemonEvolved: mothim,
+    pokemonsBeforeEvolution: wormadam
+  }: {
+    pokemonEvolved: Pokemon
+    pokemonsBeforeEvolution: Pokemon[]
+  }) {
+    const preEvolve = wormadam.pop() ?? Pokemon
+    switch (preEvolve?.name) {
+      case 'WORMADAM_SANDY':
+        mothim.types.add(Synergy.GROUND)
+        break
+      case 'WORMADAM_TRASH':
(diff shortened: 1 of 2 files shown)


app/models/colyseus-models/pokemon.ts
```diff
@@ -13772,6 +13793,27 @@ export class WormadamSandy extends Pokemon {
...
   }
+
+  onEvolve({
+    pokemonEvolved: mothim,
+    pokemonsBeforeEvolution: wormadam
+  }: {
+    pokemonEvolved: Pokemon
+    pokemonsBeforeEvolution: Pokemon[]
+  }) {
+    const preEvolve = wormadam.pop()
```
sylvainpolletvillard, line 13804: prefer non-mutative method `.at(-1)` ; `wormadams[0]` should work as well i guess

app/models/colyseus-models/pokemon.ts
```diff
@@ -13772,6 +13793,27 @@ export class WormadamSandy extends Pokemon {
...
+    switch (true) {
```
sylvainpolletvillard, line 13805: prefer a series of if/else, both shorter and more clear

app/models/colyseus-models/pokemon.ts
```diff
@@ -13798,6 +13840,27 @@ export class WormadamTrash extends Pokemon {
       !regionSynergies.includes(Synergy.GRASS)
     )
   }
+
+  onEvolve({
```
sylvainpolletvillard, line 13844: instead of duplicating code, point to a common reference. See what we did with `rksSystemOnChangePosition`

sylvainpolletvillard (approved):
lgtm for functional, just a few code remarks

PR #2301, opened 2024-10-08
Fix ghost simulation spectated

sylvainpolletvillard:
hotfix for https://discord.com/channels/737230355039387749/1292856024646156320

### DIFF
--- a/app/core/simulation.ts
+++ b/app/core/simulation.ts
@@ -226,19 +226,6 @@ export default class Simulation extends Schema implements ISimulation {
     this.applyPostEffects()
   }
 
-  getCurrentBattleResult(playerId: string) {
-    if (this.blueTeam.size === 0 && this.redTeam.size > 0) {
-      return playerId === this.bluePlayer?.id
-        ? BattleResult.DEFEAT
-        : BattleResult.WIN
-    } else if (this.redTeam.size === 0 && this.blueTeam.size > 0) {
-      return playerId === this.redPlayer?.id
-        ? BattleResult.DEFEAT
-        : BattleResult.WIN
-    }
-    return BattleResult.DRAW
-  }
-
   getEffects(playerId: string) {
     return playerId === this.bluePlayer?.id
       ? this.blueEffects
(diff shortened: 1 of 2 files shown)


app/core/simulation.ts
```diff
@@ -226,19 +226,6 @@ export default class Simulation extends Schema implements ISimulation {
     this.applyPostEffects()
   }
 
-  getCurrentBattleResult(playerId: string) {
```
sylvainpolletvillard, line 229: this method is used nowhere in the codebase so i removed it. Please double check with me

app/core/simulation.ts
```diff
@@ -1503,7 +1490,11 @@ export default class Simulation extends Schema implements ISimulation {
       })
     }
 
-    if (this.redPlayer && this.id === this.redPlayer.simulationId) {
+    if (
+      this.redPlayer &&
+      this.id === this.redPlayer.simulationId &&
+      !this.isGhostBattle
```
sylvainpolletvillard, line 1496: this prevents adding a battle result history when playing ghost. This is normally taken care by redPlayer.simulationId not being attributed to this simulation, but i prefer to double check

app/core/simulation.ts
```diff
@@ -1521,12 +1512,9 @@ export default class Simulation extends Schema implements ISimulation {
       )
 
       if (this.winnerId === this.redPlayerId) {
-        if (this.bluePlayerId !== "pve" && !this.isGhostBattle) {
-          this.redPlayer.addMoney(1, true, null)
-          client?.send(Transfer.PLAYER_INCOME, 1)
-        }
-      } else if (!this.isGhostBattle) {
-        // if blue player won against ghost, do not make red player take damage
+        this.redPlayer.addMoney(1, true, null)
```
sylvainpolletvillard, line 1515: I checked, pve team is always redPlayer, just like ghost player is always redPlayer. So this condition was not necessary

app/public/src/game/game-container.ts
```diff
@@ -70,7 +70,7 @@ class GameContainer {
   initializeSimulation(simulation: Simulation) {
     if (
       simulation.bluePlayerId === this.player?.id ||
-      simulation.redPlayerId === this.player?.id
+      (simulation.redPlayerId === this.player?.id && !simulation.isGhostBattle)
```
sylvainpolletvillard, line 73: this is the actual bug, we were displaying the wrong simulation client-side because we now preserve player.id when attributing the ghost matchup.

app/core/simulation.ts
keldaan-ag: indeed, seems to be legacy  code

app/core/simulation.ts
keldaan-ag: Indeed, seems to be legacy code

app/core/simulation.ts
keldaan-ag: 👍

PR #2310, opened 2024-10-09
added german translations for many texts

RambonL:
had to split up commits because Fink crashed often


sylvainpolletvillard:
It's fine, you're encourage to commit often to avoid risking losing progress

PR #2357, opened 2024-10-21
Partial translation of terms into Chinese

Maogee:
The pkm part is mostly an accurate translation, but the other parts may not be entirely accurate.


sylvainpolletvillard:
Thank you !

PR #2366, opened 2024-10-24
Fix clone passives

sylvainpolletvillard:
Add a onSpawn() callback on Pokemon. Contrary to afterSimulationStart, it is called on all spawns including clones/transformations into.

Fix  https://discord.com/channels/737230355039387749/1299023931658997760


keldaan-ag:
👍

PR #2404, opened 2024-11-03
Update translations (CH)

Benzene114:
All names, including the JP translations, are taken from 52Poke Wiki, the CN partner site for Bulbapedia, unless it's not available there.
Also modified some form names to make brackets more consistent between EN and CN (If EN has it then CN would, or vise versa)


sylvainpolletvillard:
Thank you !

PR #2420, opened 2024-11-07
Update translations (CH - Ability / Effect Desc Completed)

Benzene114:
This is the last batch (for now) of untranslated ability descriptions for CN.

Also edited some of my earlier translations so they're more consistent in style.


Benzene114:
The effects and their descriptions are done too

Refined some of the existing translations (mostly removing unnecessary brackets)

sylvainpolletvillard:
Thank you !

PR #2424, opened 2024-11-07
Relies on Pull Request #2422. Update ability descriptions to fit the grammar of the new status condition names in English.

Exeatop:
Includes updates for abilities that contain the following status conditions:
BURN
FREEZE
SILENCE
FATIGUE
POISONNED
PROTECT
RUNE_PROTECT
SLEEP


sylvainpolletvillard:
Hmmm... after seeing the result, i'm not really convinced. Some messages look better but some feel worse than the original. "Rune Protected" and "Armor Reduced" sound much worse for example

PR #2426, opened 2024-11-08
Update translations (ZH, JP - Item Names)

Benzene114:
Official names taken from Bulbapedia's partner sites. Non-official JP names are left blank. 

Also fixed some bad machine translations in JP names (replaced with official) and some of my previous inconsistencies.


sylvainpolletvillard:
Thank you !

PR #2427, opened 2024-11-08
Translation Conversion (ZHT → ZHS), and ZH Item Desc Translation

Benzene114:
Minor refinements to existing translations, the most significant change being changing all Traditional Chinese into Simplified for consistency's sake, as I've been translating using the latter.

If needed someone could call for a split between simplified and traditional Chinese, like how the mainline Pokemon games have CHS and CHT as separate language, but for now I'd treat ZH as simplified.


Benzene114:
And item descriptions are all translated

sylvainpolletvillard:
Thank you !

PR #2435, opened 2024-11-10
Fix AP scaling on some abilities

Phoenix-Starlight:
Diamond Storm and Night Shade both apply AP before-hand.

Their respective AP scaling as of now is `AP^2 + 2AP + 1` and `AP^2/2 + 1.5AP + 1`. At 100AP this means 4x and 3x respectively.

Trick Or Treat incorrectly applies AP on one of its curses.

### DIFF
--- a/app/core/abilities/abilities.ts
+++ b/app/core/abilities/abilities.ts
@@ -10394,13 +10394,13 @@ export class TrickOrTreatStrategy extends AbilityStrategy {
       target.hp = Math.floor(target.hp * lifeReduction)
       target.status.triggerFlinch(3000, target, pokemon)
     } else if (pokemon.ap <= 100) {
-      // 51-100 AP: transforms the unit in magikarp during X seconds, replacing its ability with splash
+      // 51-100 AP: transforms the unit into magikarp for X seconds, replacing its ability with splash
       const originalAbility = target.skill
       const originalAttack = target.atk
       const originalDefense = target.def
       const originalSpecialDefense = target.speDef
       const originalIndex = target.index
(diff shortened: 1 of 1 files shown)


app/core/abilities/abilities.ts
```diff
@@ -10388,12 +10389,12 @@ export class TrickOrTreatStrategy extends AbilityStrategy {
       }
     } else if (pokemon.ap <= 50) {
       // 0-50 AP: shrink unit size and HP
-      const lifeReduction = 0.4 * (1 + pokemon.ap / 100)
+      const lifeReduction = 0.4 * (1 - pokemon.ap / 100)
```
sylvainpolletvillard, line 10392: this is incorrect, it could lead to negative life reduction. The correct formula intended is 
```suggestion
      const lifeReduction = 0.4 / (1 + pokemon.ap / 100)
```

app/core/abilities/abilities.ts
```diff
@@ -761,7 +761,8 @@ export class DiamondStormStrategy extends AbilityStrategy {
           board,
           AttackType.SPECIAL,
           pokemon,
-          crit
+          crit,
+          false
```
sylvainpolletvillard, line 765: i'd rather remove the ap part in damage compute above

sylvainpolletvillard (changes requested):

app/core/abilities/abilities.ts
Phoenix-Starlight: It's not possible regardless since it will pick a different curse above 50AP, but will do.

sylvainpolletvillard:
Thank you !

PR #2440, opened 2024-11-10
Fixed quadratically scaling crit chance on some abilities.

Exeatop:
Sky Attack Shadow and Night Slash can crit by default. Their crit calculation is applied twice by including `crit ||` which gives a crit probability scaling quadratically with crit chance if the user is holding Reaper Cloth. Removing `crit ||` fixes the issue.


Exeatop:
> Intended, it only buffs crit chance when reaper cloth is equipped, and give a good reason to equip reaper cloth on these units

Oh is this intended? If that's the case then the other abilities that crit by default should have it added since they're currently missing the reaper cloth crit chance.

PR #2445, opened 2024-11-13
Ghost Battle abilities duplication fix

Exeatop:
Fixes Ghost Battle duplication for the following abilities:

Payday
Pickup
Knowledge Thief
Eggsplosion
Gold Rush


sylvainpolletvillard:
This test does not check if the player is red team or blue team. Blue team in a ghost matchup still needs to get these bonuses

Also we already have a check in addMoney() function for ghost matchups. A similar check should be done for xp gains and unit gains

PR #2451, opened 2024-11-15
Fix auto ready race condition

Phoenix-Starlight:
Due to joins being async, it can be ordered after the ready command, causing the ready command to fail.
Additionally, prior to the fix, the ready button did not work at all in non-custom lobbies since `undefined !== true` would cause an early return.

### DIFF
--- a/app/rooms/commands/preparation-commands.ts
+++ b/app/rooms/commands/preparation-commands.ts
@@ -5,7 +5,7 @@ import { FilterQuery } from "mongoose"
 import { GameUser, IGameUser } from "../../models/colyseus-models/game-user"
 import { BotV2, IBot } from "../../models/mongo-models/bot-v2"
 import UserMetadata from "../../models/mongo-models/user-metadata"
-import { Role } from "../../types"
+import { Role, Transfer } from "../../types"
 import {
   EloRank,
   EloRankThreshold,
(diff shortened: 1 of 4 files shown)


app/rooms/commands/preparation-commands.ts
```diff
@@ -144,6 +132,9 @@ export class OnJoinCommand extends Command<
           )
         }
       }
+      if (this.room.state.gameMode !== GameMode.CUSTOM_LOBBY) {
```
sylvainpolletvillard, line 135: This won't work. The whole point of the ready check is to make sure that user has loaded the preparation room and has the event listeners ready to start game when the START_GAME message is sent. Otherwise I would have just removed the ready system completely if it was entirely managed by the server

app/rooms/commands/preparation-commands.ts
```diff
@@ -549,6 +540,7 @@ export class OnToggleReadyCommand extends Command<
   execute({ client, ready }) {
     try {
       // cannot toggle ready in quick play / ranked / tournament game mode
+      if (this.room.state.gameMode !== GameMode.CUSTOM_LOBBY && ready !== true && ready !== undefined)
```
sylvainpolletvillard, line 543: why these 2 consecutive ifs ?

Phoenix-Starlight: the ready button sends `undefined` which is why i added the check.

Phoenix-Starlight: I suppose it can be removed.

app/rooms/commands/preparation-commands.ts
Phoenix-Starlight: Ok, I see.

app/rooms/commands/preparation-commands.ts
Phoenix-Starlight: Wait oh. That's an accidental line addition.

Phoenix-Starlight:
Unfortunately, immediately sending the message sometimes results in it not being received.

sylvainpolletvillard:
I need you to explain your reasoning here. As far as I understand these changes, the server is now sending to every client joining every 250ms a TOGGLE_READY message that ask him to send a toggleReady like a ping/pong exchange. How is that necessary, and why do you think it will solve the problem ?

sylvainpolletvillard:
related: https://github.com/keldaanCommunity/pokemonAutoChess/pull/2455

Phoenix-Starlight:
If the client is not replying with `toggle_ready(true)` that means the client is not ready, or the client is not responding.
The server has to ask the client to ready up because the DB call can return at any point in time.

Previously:
- Server receives join command and fetches from DB
- Client tries toggling ready
- Server processes it but the client isn't in `state.users` and doesn't ready the user
- DB call returns and sets the kick timer
- User gets kicked

Now:
- Server receives join command and fetches from DB
- Client sets a listener
- Server sends a message at an interval, or doesn't because there are more than 8 players. The kick timer is set beforehand.
- Client responds back
- Server readies the user and cancels both timers.

Even if the client sets a listener after the first message is sent, it will receive the second one.

sylvainpolletvillard:
DB as database ? What are you referring to ?

PR #2460, opened 2024-11-17
Zygarde

Windchili:
Based on NerdySatyr's design, with this change:


keldaan-ag:
rebase and we're good

app/core/pokemon-entity.ts
```diff
@@ -1251,6 +1251,29 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
...
+        this.addDefense(1, this, 0, false)
+        this.addSpecialDefense(1, this, 0, false)
+        this.addPP(20, this, 0, false)
+        this.addMaxHP(50, this, 0, false)
+      } else {
+        this.addAttack(5, this, 0, false)
+        this.addDefense(2, this, 0, false)
+        this.addSpecialDefense(2, this, 0, false)
+        this.addMaxHP(80, this, 0, false)
+        this.range = 1
```
sylvainpolletvillard, line 1267: need to take into account wide lens. Use range = min(1)(range - difference) instead

sylvainpolletvillard (changes requested):

Windchili (commented):
Just looking at the code, everything seems good and I agree with the changes suggested. Thanks for reformatting it to look nicer.

PR #2487, opened 2024-11-24
Fix Lock On missing

Phoenix-Starlight:
Oops.

### DIFF
--- a/app/core/pokemon-state.ts
+++ b/app/core/pokemon-state.ts
@@ -68,23 +68,6 @@ export default abstract class PokemonState {
         damage += Math.round(pokemon.def * (1 + pokemon.ap / 100))
       }
 
-      let trueDamagePart = 0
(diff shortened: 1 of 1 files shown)


app/core/pokemon-state.ts
```diff
@@ -104,10 +103,11 @@ export default abstract class PokemonState {
       }
       dodgeChance = max(0.9)(dodgeChance)
 
-      if (
+      if (pokemon.effects.has(Effect.LOCK_ON)) {
+        pokemon.effects.delete(Effect.LOCK_ON)
```
Exeatop, line 107: As noted in the other PR, this fix introduces a bug where the effect will be deleted if the target is lost. Just move the trueDamagePart calculation to be directly above the conditional if(trueDamagePart > 0)

Phoenix-Starlight: Oops.

PR #2488, opened 2024-11-24
Hotfix Lock On missing

Phoenix-Starlight:


app/core/pokemon-state.ts
```diff
@@ -104,10 +103,11 @@ export default abstract class PokemonState {
       }
       dodgeChance = max(0.9)(dodgeChance)
 
-      if (
+      if (pokemon.effects.has(Effect.LOCK_ON)) {
+        pokemon.effects.delete(Effect.LOCK_ON)
```
Exeatop, line 107: This fix introduces a bug where the effect will be deleted if the target is lost. Just move all of the trueDamagePart calculations to be directly above the conditional if(trueDamagePart > 0)

PR #2490, opened 2024-11-26
Fix Outrage and Torch Song attacking their targets twice

Exeatop:


sylvainpolletvillard:
This fix is incorrect. it should target all adjacent enemies AND at least the current target if not adjacent

PR #2499, opened 2024-11-29
Fix ice synergy freeze chance

Exeatop:


sylvainpolletvillard:
Oopsie. Big bug here, good catch

PR #2511, opened 2024-12-01
Fix permanent adjustments in ghost battles

Exeatop:
Added two new methods to PokemonEntity:
removePermanentItem and addPermanentItem
Also refactored the applyStat method to handle changes to refToBoardPokemon
These methods allow for adjusting stats and items without directly interfacing with instance variables.
This way bugs can be fixed in all instances of an issue, such as checking for isGhostOpponent in this case.


app/core/pokemon-entity.ts
```diff
@@ -535,6 +535,28 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
...
+      }[keyof IPokemon],
+    value: number
+  ) {
+    if(this.isGhostOpponent) return
+    this.refToBoardPokemon[stat] += value
+  }
+
+  addPermanentItem(
+    item: Item,
+    remove = false
```
sylvainpolletvillard, line 551: confusing method name for removing, declare a `removePermanentItem` method instead

app/core/pokemon-entity.ts
```diff
@@ -535,6 +535,28 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
...
   }
+  
+  addPermanentStats(
+    stat: { [K in keyof IPokemon]:
+      IPokemon[K] extends number ? K : never
+      }[keyof IPokemon],
+    value: number
+  ) {
+    if(this.isGhostOpponent) return
+    this.refToBoardPokemon[stat] += value
```
sylvainpolletvillard, line 546: I suggest refactoring the `applyStat` function so that it takes a `permanent` boolean flag, this way we can reuse all the value-checks / rounding / min / max of all the existing stat buffs methods in PokemonEntity

app/core/pokemon-entity.ts
```diff
@@ -1404,7 +1426,7 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
       target.status.triggerProtect(2000)
       target.handleHeal(20, target, 0, false)
       target.items.delete(Item.BABIRI_BERRY)
-      target.refToBoardPokemon.items.delete(Item.BABIRI_BERRY)
+      target.addPermanentItem(Item.BABIRI_BERRY, true)
```
sylvainpolletvillard, line 1429: the `removePermanentItem` method could handle removing the item both from the entity and the board pokémon

app/core/pokemon-entity.ts
```diff
@@ -1547,7 +1569,7 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
 
   // called after death (does not proc if resurection)
   onDeath({ board }: { board: Board }) {
-    this.refToBoardPokemon.deathCount++
+    this.addPermanentStats("deathCount", 1)
```
sylvainpolletvillard, line 1572: confusing method name here, this is not a stat. I'd rather keep original code with a if condition for ghosts

sylvainpolletvillard (changes requested):
Good initiative, but I'd rather reuse the existing stat buffs methods. Suggested a few refactoring

app/core/pokemon-entity.ts
Exeatop: Yeah that's a much better idea. Should've thought about that first.

Exeatop: Oh I think my change isn't what you wanted. Let me see about doing the full refactor

Exeatop: ok should be good now

app/core/abilities/abilities.ts
```diff
@@ -5527,11 +5527,9 @@ export class FellStingerStrategy extends AbilityStrategy {
       pokemon.addAbilityPower(5, pokemon, 0, false)
       pokemon.addAttack(1, pokemon, 0, false)
       pokemon.addMaxHP(10, pokemon, 0, false)
-      if (!pokemon.isGhostOpponent) {
-        pokemon.refToBoardPokemon.atk += 1
-        pokemon.refToBoardPokemon.ap += 5
-        pokemon.refToBoardPokemon.hp += 10
-      }
+      pokemon.applyStat(Stat.ATK, 1, true)
```
sylvainpolletvillard, line 5530: should just be

```
    pokemon.addAbilityPower(5, pokemon, 0, false, true)
      pokemon.addAttack(1, pokemon, 0, false, true)
      pokemon.addMaxHP(10, pokemon, 0, false, true)
```

app/core/pokemon-entity.ts
```diff
@@ -435,11 +435,14 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
...
-    crit: boolean
+    crit: boolean,
+    permanent = false
   ) {
     value =
       value * (1 + (apBoost * caster.ap) / 100) * (crit ? caster.critPower : 1)
-    this.hp = min(1)(this.hp + value)
+    const target = permanent ? this.refToBoardPokemon : this
+    target.hp = min(1)(target.hp + value)
+    if(permanent) return
```
sylvainpolletvillard, line 445: you forget to check for ghost opponents which was the initial point of this PR
Also there's no scenario where we want to add a stat on the board pokemon but not the entity, so there's no point for an early return here

sylvainpolletvillard:
You missed my point, I suggested to add the permanent behavior in addition to the in-fight behavior. There's no scenario where we want the permanent buff to not apply on the entity mid-fight as well.

sylvainpolletvillard:
Good ! If I nitpick i would do `addItem(item: Item, permanent = false)`, but that's already goodenough

PR #2538, opened 2024-12-05
allow addMaxHP and addDodgeChance methods to properly handle negative values

Exeatop:


app/core/pokemon-entity.ts
```diff
@@ -444,7 +444,11 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
       target.hp = min(1)(target.hp + value)
     }
     update(this)
-    this.life = max(this.hp)(this.life + value)
+    if (value > 0) {
+      this.life = max(this.hp)(this.life + value)
+    } else {
+      this.life = max(this.hp)(this.life)
```
sylvainpolletvillard, line 450: this change doesn't make sense to me, if a unit lose max HP it should lose life as well

sylvainpolletvillard (changes requested):
the change for maxHP is incorrect for me; Dodge OK

PR #2540, opened 2024-12-05
new method simulation.removeItemEffect

Exeatop:
I'm not entirely sure it makes sense to have this be a method for simulation when it's applied only by a PokemonEntity object, but keeping it in the same place as applyItemEffect probably makes it easier to organize mentally.


sylvainpolletvillard:
agree it doesn't really make sense in simulation class. A good way to find out this is when the method doesn't use the `this` keyword at all

PR #2542, opened 2024-12-06
Fix fallen allies calculation for Kowtow Cleave and Retaliate abilities

Exeatop:


sylvainpolletvillard:
How about using `blueTeam.size` instead of `blueDpsMeter.size` ? Team is also updated with `addPokemon` and it feels more natural than checking against a dps meter

Exeatop:
> How about using `blueTeam.size` instead of `blueDpsMeter.size` ? Team is also updated with `addPokemon` and it feels more natural than checking against a dps meter

I agree that checking against a dps meter is a bit odd, but it's the only solution I was able to find without adding additional features. Checking for blueTeam.size doesn't work because we need to know how many allied pokemon have been at play in total, but they are deleted from the team during simulation.update if they are dead:

 if (
        (!pkm.life || pkm.life <= 0) &&
        !pkm.status.resurecting &&
        !pkm.status.resurection
      ) {
        this.blueTeam.delete(key)

PR #2543, opened 2024-12-06
Update translations (ZH - Latest + ZH - Fixes + Misc - Name Fixes)

Benzene114:
fixes some format errors and improved a few translations in ZH, in addition to adding the latest additions.

Plus the still ongoing official name list filling.


app/public/dist/client/locales/fr/translation.json
```diff
@@ -1163,12 +1163,12 @@
     "WATTREL": "Zapétrel",
     "KILOWATTREL": "Fulgulairo",
     "STANTLER": "Cerfrousse",
-    "BURMY_PLANT": "Cheniti (Cape Plante)",
-    "BURMY_SANDY": "Cheniti (Cape Sable)",
-    "BURMY_TRASH": "Cheniti (Cape Déchet)",
-    "WORMADAM_PLANT": "Cheniselle (Cape Plante)",
-    "WORMADAM_SANDY": "Cheniselle (Cape Sable)",
-    "WORMADAM_TRASH": "Cheniselle (Cape Déchet)",
+    "BURMY_PLANT": "Cheniti (Plante)",
```
sylvainpolletvillard, line 1166: Cape Plante is the right name of this form, I'm sure of that because it's my language

sylvainpolletvillard (changes requested):
When changing other languages, please double check. I'd rather you focus on the languages you speak instead of assuming things from wikis

app/public/dist/client/locales/fr/translation.json
Benzene114: Checked through wikis as stated in commits; Lycanroc's form names stay changed because the sites say so

Benzene114: honestly this is more of a "overthinking" moment as EN lacks the "Cloak" in brackets (corresponds to "Cape" in FR and "Manto" in ES), my bad

PR #2544, opened 2024-12-06
Refactor simulation.applyItemEffect

Exeatop:
moved item application logic to core/items.ts
- allows for item logic to share the same hash table
    - item removal logic coming next
- isn't simulation specific (not needed to be a simulation method)

added EffectClass type for function hooks

updated simulation.applyItemEffect calls


sylvainpolletvillard:
This is declaring 16 functions and a map everytime you call `triggerItem`. Very inefficient.

I don't see how this refactoring can be a good thing, for now all i can see is more lines of code, less clarity and less performance

PR #2548, opened 2024-12-08
Fix 9 out of 8 rooms

Phoenix-Starlight:
Rejecting in `onJoin` will cause the same result as double joins, where the user becomes a ghost of the room.
This fix will trigger warnings and errors for onAuth failing on the client's console upon rejection.
As well, Colyseus is updated to include a fix for `onAuth` rejections not decrementing client count.

### DIFF
--- a/app/rooms/preparation-room.ts
+++ b/app/rooms/preparation-room.ts
@@ -1,5 +1,5 @@
 import { Dispatcher } from "@colyseus/command"
-import { Client, Room, updateLobby } from "colyseus"
+import { Client, ClientArray, Room, updateLobby } from "colyseus"
 import admin from "firebase-admin"
 import BannedUser from "../models/mongo-models/banned-user"
 import { IBot } from "../models/mongo-models/bot-v2"
@@ -28,9 +28,11 @@ import {
   RemoveMessageCommand
 } from "./commands/preparation-commands"
 import PreparationState from "./states/preparation-state"
+import { UserRecord } from "firebase-admin/lib/auth/user-record"
 
 export default class PreparationRoom extends Room<PreparationState> {
   dispatcher: Dispatcher<this>
+  clients!: ClientArray<undefined, UserRecord>
 
   constructor() {
     super()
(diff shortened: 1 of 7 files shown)


app/utils/function.ts
```diff
@@ -44,6 +44,27 @@ export function throttle<T extends (...args: any) => any>(
...
+export function block<T extends (...args: any) => any>(
+  fn: T
+): ThrottledFunction<T> {
+  let executing: boolean
+  let lastResult: ReturnType<T>
+  
+  return async function (this: any, ...args: any[]): Promise<ReturnType<T>> {
+    const context = this
+    
+    if (!executing) {
```
sylvainpolletvillard, line 56: I don't know what this function is supposed to do but i'm pretty sure it's bugged, since calling this function twice in a row will set executing to false no matter the timing of the asynchronous task

app/rooms/preparation-room.ts
```diff
@@ -337,18 +339,23 @@ export default class PreparationRoom extends Room<PreparationState> {
...
       const userProfile = await UserMetadata.findOne({ uid: user.uid })
       client.send(Transfer.USER_PROFILE, userProfile)
 
-      const isAlreadyInRoom = this.state.users.has(user.uid)
       const numberOfHumanPlayers = values(this.state.users).filter(
         (u) => !u.isBot
       ).length
-      if (numberOfHumanPlayers >= MAX_PLAYERS_PER_GAME && !isAlreadyInRoom) {
+
+      if (!this.state.uids.has(user.uid)) {
```
sylvainpolletvillard, line 352: this `uids` should not be needed, you can just check `state.users.some(u => u.uid === user.uid)`

sylvainpolletvillard (changes requested):

app/utils/function.ts
Phoenix-Starlight: Good catch

app/rooms/preparation-room.ts
Phoenix-Starlight: `this.state.users` is only updated in `OnJoinCommand`.

sylvainpolletvillard: doesn't change my point, to check if the user is already there, it should already have gone through the onjoin command

Phoenix-Starlight: No, onAuth comes before onJoin.

Phoenix-Starlight: Ok, I stand corrected after some testing.

sylvainpolletvillard: I know. But you're testing if the user has **already** joined here. Which means he is in state.users

app/utils/function.ts
sylvainpolletvillard: You may want to add a comment to clarify what the function does. It's a bit strange to me because usually if i want an async function to be called once and return the appropriate result, i store the promise, not the lastResult

PR #2560, opened 2024-12-10
maractus rework

sylvainpolletvillard:
Maractus is a maraccas, it should be sound type. Plus it also gives more variety to sound compositions which is needed. Also makes it worth to cast again Spike Armor while the previous effect is not depleted.


keldaan-ag:
👍

PR #2569, opened 2024-12-12
Add item effect removal logic

Exeatop:
"What happens when an item is removed mid-fight" from: https://discord.com/channels/737230355039387749/1302243528088486011/1302243528088486011


Exeatop:
I'll adjust formatting and add comments about the intentionally missing removal effects after this goes through.

sylvainpolletvillard:
Nice work 👍

PR #2578, opened 2024-12-13
Update synergy effects with items

Exeatop:
- add stat gains from added synergies
- add dynamic effects from added synergies
- remove dynamic effects from lost synergies
- update patch notes to be aligned with previous item changes

https://discord.com/channels/737230355039387749/1302243528088486011/1302243528088486011


app/core/pokemon-entity.ts
```diff
@@ -591,11 +591,45 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
...
+
     this.items.add(item)
     this.simulation.applyItemEffect(this, item)
     if (permanent && !this.isGhostOpponent) {
       this.refToBoardPokemon.items.add(item)
     }
+
+    const addSynergyFromItem = () => {
+      this.types.add(type)
+      if (type === Synergy.GHOST) {
```
sylvainpolletvillard, line 607: We need to find a way to reuse `simulation.applySynergyEffects` here, in order to avoid duplicating all the logic. How about a second `synergy` parameter that filters the SynergyEffects applied ?

app/core/pokemon-entity.ts
```diff
@@ -591,11 +591,45 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
   }
 
   addItem(item: Item, permanent = false) {
+    const type = SynergyGivenByItem[item]
+    if (type && this.types.has(type)) {
+      return // prevent adding synergy items for existing types
```
sylvainpolletvillard, line 596: about that, we're going to make some changes

image


for context, the initial prevention for synergy stones was added at a later stage to avoid a footgun for the players. But artificial items can be moved at any time and have other bonuses that just their synergy.

app/core/pokemon-entity.ts
Exeatop: I refactored applySynergyEffects for this and it's so much better

app/core/pokemon-entity.ts
Exeatop: allowing duplicate types means that if the item is removed, synergy effects can be removed even though they shouldn't since they still have that type. I couldn't find an easy way to check for this, so I think it should be addressed in another PR.

PR #2592, opened 2024-12-16
Prevent selecting/dragging button text

Phoenix-Starlight:
Currently one can select or drag content within the modal for creating/joining games. This disables being able to select content within the buttons for the modal.


sylvainpolletvillard:
You can use `user-drag: none` in CSS to do the same. This kind of behavior could be applied to lots of UI elements, if not all

Phoenix-Starlight:
https://caniuse.com/?search=user-drag

There is no non-prefixed css property.

The draggable attribute is supported universally.

PR #2610, opened 2024-12-20
keep shiny when evolving with shiny charm

sylvainpolletvillard:
https://discord.com/channels/737230355039387749/1319695402211213424


Exeatop:
This fix only considers pokemon that evolve by count. It ignores other methods of evolution and transformations. Adding it to `carryOverPermanentStats` should cover every case.

sylvainpolletvillard:
ok works for me

PR #2614, opened 2024-12-21
Fix resurrect losing and duplicating some effects

Exeatop:
https://discord.com/channels/737230355039387749/1317095176333824041

### DIFF
--- a/app/core/pokemon-entity.ts
+++ b/app/core/pokemon-entity.ts
@@ -19,6 +19,7 @@ import {
   ARMOR_FACTOR,
   BOARD_HEIGHT,
   BOARD_WIDTH,
+  DEFAULT_ATK_SPEED,
   DEFAULT_CRIT_CHANCE,
   DEFAULT_CRIT_POWER,
   ItemStats,
@@ -1762,16 +1763,26 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
     this.life = this.refToBoardPokemon.hp
     this.shield = 0
     this.pp = 0
-    this.ap = 0
+    this.ap = this.refToBoardPokemon.ap
     this.atk = this.refToBoardPokemon.atk
     this.def = this.refToBoardPokemon.def
     this.speDef = this.refToBoardPokemon.speDef
     this.atkSpeed = this.refToBoardPokemon.atkSpeed
     this.critChance = DEFAULT_CRIT_CHANCE
     this.critPower = DEFAULT_CRIT_POWER
+    this.atkSpeed = DEFAULT_ATK_SPEED
+    this.range = this.refToBoardPokemon.range
+    this.dodge = 0
+    this.luck = 0
     this.count = new Count()
-    this.status.clearNegativeStatus()
     this.effects.clear()
+    this.status = new Status({
+      enrageDelay: this.status.enrageDelay,
+      curseVulnerability: this.status.curseVulnerability,
+      curseWeakness: this.status.curseWeakness,
+      curseTorment: this.status.curseTorment,
+      curseFate: this.status.curseFate
+    })
 
     if (this.items.has(Item.SACRED_ASH) && this.player) {
       const team = this.simulation.getTeam(this.player.id)
@@ -1816,6 +1827,11 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
     ) // prevent resurecting fossils twice
 
     // does not trigger postEffects (iron defense, normal shield, rune protect, focus band, delta orb, flame orb...)
+
+    this.refToBoardPokemon.onSpawn({
+      entity: this,
+      simulation: this.simulation
+    })
   }
 
   eatBerry(berry: Item, stealedFrom?: PokemonEntity) {
--- a/app/core/pokemon-entity.ts
+++ b/app/core/pokemon-entity.ts
@@ -19,7 +19,6 @@ import {
   ARMOR_FACTOR,
   BOARD_HEIGHT,
   BOARD_WIDTH,
-  DEFAULT_ATK_SPEED,
   DEFAULT_CRIT_CHANCE,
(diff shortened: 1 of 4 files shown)


app/core/pokemon-entity.ts
```diff
@@ -1762,16 +1763,26 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
...
     this.pp = 0
-    this.ap = 0
+    this.ap = this.refToBoardPokemon.ap
     this.atk = this.refToBoardPokemon.atk
     this.def = this.refToBoardPokemon.def
     this.speDef = this.refToBoardPokemon.speDef
     this.atkSpeed = this.refToBoardPokemon.atkSpeed
     this.critChance = DEFAULT_CRIT_CHANCE
     this.critPower = DEFAULT_CRIT_POWER
+    this.atkSpeed = DEFAULT_ATK_SPEED
```
sylvainpolletvillard, line 1773: this is assigned twice. Why not copying the board pokemon value here ?

app/core/pokemon-entity.ts
```diff
@@ -1762,16 +1763,26 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
...
+    this.range = this.refToBoardPokemon.range
+    this.dodge = 0
+    this.luck = 0
     this.count = new Count()
-    this.status.clearNegativeStatus()
     this.effects.clear()
+    this.status = new Status({
```
sylvainpolletvillard, line 1779: should probably be a clear() method instead of losing the existing instance. Why not `clearNegativeStatus` ?

app/core/pokemon-entity.ts
```diff
@@ -1816,6 +1827,11 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
     ) // prevent resurecting fossils twice
 
     // does not trigger postEffects (iron defense, normal shield, rune protect, focus band, delta orb, flame orb...)
+
+    this.refToBoardPokemon.onSpawn({
```
sylvainpolletvillard, line 1831: I don't think resurrect() should proc spawn effects. You dont want Sudowoodo to become a tree again,

sylvainpolletvillard (changes requested):

app/core/pokemon-entity.ts
Exeatop: Status isn't a set so it can't be directly cleared. `clearNegativeStatus` doesn't encompass all of the effects that need to be cleared. There are far fewer properties that need to be retained than those that need to be reset, so it's easier to retain those fewer properties.

app/core/pokemon-entity.ts
Exeatop: `resurrect()` not proccing spawn effects causes a few bugs. Trees seem to be the only case where it's undesirable to have them trigger again. It's probably best to just use a conditional then.

sylvainpolletvillard: What bugs ? It's the first that came when looking up for onSpawn callbacks. Regigigas -25% attack speed is not desirable neither.

resurrecting doesn't mean spawning a new unit, therefore it should not call onSpawn effects. As simple as that.

app/core/pokemon-entity.ts
sylvainpolletvillard: I'm not referring to a Set, i'm referring to declaring a method to reset the statuses properly on resurrect. I also believe its for the best to manually specify which statuses should be reset on resurrection, because we may want to preserve some. For example, the enraged status should not be reset after 35 seconds of fight. Other passive-linked statuses like Clear Wing should not be cleared neither.

app/models/colyseus-models/pokemon.ts
```diff
@@ -179,6 +179,7 @@ export class Pokemon extends Schema implements IPokemon {
   onSpawn(params: {
     entity: IPokemonEntity
     simulation: Simulation
+    resurrect?: boolean
```
sylvainpolletvillard, line 182: This doesn't make sense to me

sylvainpolletvillard (changes requested):

app/core/pokemon-entity.ts
Exeatop: Another word for resurrect is respawn, hence spawn. It's wholly unintuitive that effects which aren't explicitly stated as triggering "at the start of the fight" would not be reset with everything else on resurrection. Why should Regigigas' passive not be reapplied? Currently its attack speed goes like this: 50 -> 80 -> (resurrect) -> 75 -> 105. So even if you wanted to not reapply its passive, its implementation needs to be changed because this is certainly unintended.

app/core/pokemon-entity.ts
Exeatop: How is Clear Wing any different from Ground stacks or Soul Dew? It's not obvious why passives shouldn't be reset. 

I thought that enraged was fine to be reset because it would just be triggered again, but maybe with update delay it's not? I think it's fine to specify what to reset instead of specifying what to keep, but here's my reasoning for why I decided to specify what to keep:
The vast majority of statuses interact with only the pokemon they are attached to, so they can be reset on death. However, there are a few statuses which, as a special case, interact with the board state in some way. Ghost curses need to transmit to other pokemon and enraged needs to trigger based on the round timer. Because these cases are the exception and not the norm, it makes more sense that they are the ones explicitly stated.

app/core/pokemon-entity.ts
sylvainpolletvillard: Disagree, respawn would imply the entity has been removed and readded, which is not the case for PAC, the entity is always here with a dedicated status. onSpawn is used to specify effects triggering after the entity is created, being at the start of the fight or with a spawning ability like Houndour/Banette. Regigigas passive should not be reapplied, it is designed to be applied once and nothing indicated that it should be reset on resurrection.

PR #2616, opened 2024-12-24
Effect refactoring: grow ground timer

sylvainpolletvillard:
This is a WIP refactoring as discussed in #dev channel

This PR enables the removal of those fields:
- entity.growGroundTimer
- entity.count.growGroundCount


app/core/effect.ts
```diff
@@ -1,13 +1,68 @@
...
+
+  constructor(
+    effect: (entity: PokemonEntity) => void,
+    intervalMs: number,
+    origin?: EffectOrigin
+  ) {
+    super(effect, origin)
+    this.intervalMs = intervalMs
+    this.timer = intervalMs
+    this.count = 0
```
Exeatop, line 34: If the goal is to reduce unnecessary static property duplication, I think this should be moved to the effects that use it since not every periodic effect needs a count. It's incremented at the same time the constructed method is called, so it should be fine to use only on classes that need it.

app/core/simulation.ts
```diff
@@ -1142,6 +1142,14 @@ export default class Simulation extends Schema implements ISimulation {
...
         if (types.has(Synergy.GROUND)) {
           pokemon.effects.add(effect)
+          const synergyLevel =
+            [
+              Effect.TILLER,
+              Effect.DIGGER,
+              Effect.DRILLER,
+              Effect.DEEP_MINER
+            ].indexOf(effect) + 1
+          pokemon.effectsSet.add(new GrowGroundEffect(synergyLevel, effect))
```
Exeatop, line 1152: This array already exists in SynergyEffects[Synergy.GROUND] which I think makes more sense to use, but isn't super important.

app/core/simulation.ts
```diff
@@ -1142,6 +1142,14 @@ export default class Simulation extends Schema implements ISimulation {
       case Effect.DEEP_MINER:
         if (types.has(Synergy.GROUND)) {
           pokemon.effects.add(effect)
+          const synergyLevel =
+            [
+              Effect.TILLER,
+              Effect.DIGGER,
+              Effect.DRILLER,
+              Effect.DEEP_MINER
+            ].indexOf(effect) + 1
```
Exeatop, line 1151: This array already exists in SynergyEffects[Synergy.GROUND] which I think makes more sense to use, but isn't super important.

app/core/effect.ts
sylvainpolletvillard: Maybe, i hesitated. I could see it useful to have on generic PeriodicEffect

Exeatop: Why not extend periodic effects to another abstract class that contains countable periodic effects? That way you prevent code duplication and also prevent unnecessary properties being on some classes.

sylvainpolletvillard: felt like overkill

PR #2630, opened 2024-12-29
Set timer before auto-starting matches

Phoenix-Starlight:
Needs testing.

This pull request aims to resolve an issue where a player joins an auto-start room, then immediately leaves which occasionally causes games to start with the departing player included.

### DIFF
--- a/app/rooms/commands/preparation-commands.ts
+++ b/app/rooms/commands/preparation-commands.ts
@@ -621,14 +621,11 @@ export class OnToggleReadyCommand extends Command<
 
 export class CheckAutoStartRoom extends Command<PreparationRoom, void> {
   async execute() {
-    await setTimeout(3000)
+    try {
+      this.state.abortOnPlayerLeave = new AbortController()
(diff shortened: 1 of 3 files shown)


sylvainpolletvillard:
Code seems okay, testing right now

sylvainpolletvillard:
It didn't work, game started with 7 players. Not sure about the onRemove() callback. I'm also concerned the callback can get registered several times. 

How about checking if still 8 players after 3 seconds ?

sylvainpolletvillard:
I tested again and it didn't work again

nbExpectedPlayers 8
this.state.users.size 8

The way i test: i open 8 different firefox profiles and start a scribble. With the 8th player i immediately leave the lobby after joining. I can see the logs with "user leaved" happening before the start log. Maybe it has something to do with the allowed reconnection time for preparation room ?

Phoenix-Starlight:
Believe I ironed all the bugs out.

sylvainpolletvillard:
Yes, this time it worked ! Good work

PR #2636, opened 2025-01-02
updated DE translations

yVieta:
Fixed translations. Some translations are not correct in their meaning which also have form (expression) issues.


sylvainpolletvillard:
Danke schön !

PR #2650, opened 2025-01-05
Fix regional addpicks displaying despite not being in pool

Exeatop:
Should be double checked to see if the shop update is sufficient - I think it is though.
https://discord.com/channels/737230355039387749/1318630909347500063

### DIFF
--- a/app/public/src/pages/component/synergy/synergy-detail-component.tsx
+++ b/app/public/src/pages/component/synergy/synergy-detail-component.tsx
@@ -49,7 +49,9 @@ export default function SynergyDetailComponent(props: {
     .map((p) => getPokemonData(p as Pkm))
     .sort((a, b) => RarityCost[a.rarity] - RarityCost[b.rarity])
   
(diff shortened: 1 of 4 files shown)


app/public/src/pages/component/synergy/synergy-detail-component.tsx
```diff
@@ -48,11 +48,21 @@ export default function SynergyDetailComponent(props: {
     )
     .map((p) => getPokemonData(p as Pkm))
     .sort((a, b) => RarityCost[a.rarity] - RarityCost[b.rarity])
+  
+  const baseVariant = (basePkm: Pkm): Pkm => {
```
sylvainpolletvillard, line 52: the naming of variables is weird, i suggest:

```suggestion
  const baseVariant = (pkm: Pkm): Pkm =>  (Object.keys(PkmRegionalVariants) as Pkm[]).find((p) =>
      PkmRegionalVariants[p]!.includes(pkm)
    ) ?? pkm
```

PR #2679, opened 2025-01-10
Hotfix untargetable Timburr

Exeatop:
https://discord.com/channels/737230355039387749/1325809988177887252

### DIFF
--- a/app/core/abilities/abilities.ts
+++ b/app/core/abilities/abilities.ts
@@ -11090,6 +11090,7 @@ export class ColumnCrushStrategy extends AbilityStrategy {
         pillar.team === Team.BLUE_TEAM
           ? pillar.simulation.blueTeam
           : pillar.simulation.redTeam
+      pillar.shield = 0
       pillar.handleSpecialDamage(
         9999,
         board,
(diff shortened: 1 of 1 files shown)


sylvainpolletvillard:
This could work but also mess up the DPS report and could have weird unintended effects ? No synergy/item for the pillar so less likely, but yet... Is there anything in the handleDamage death code that we miss here ?

Exeatop:
> This could work but also mess up the DPS report and could have weird unintended effects ? No synergy/item for the pillar so less likely, but yet... Is there anything in the handleDamage death code that we miss here ?

I checked the DPS report and it's unaffected. That's the reason I set attacker to null. The only difference this introduces, as far as I can tell, is that if it was hit with a ghost curse then it would've been lost before. Now the curse transfers to the team.

sylvainpolletvillard:
We should probably exclude inanimate objects from ghost curse anyway. Why is this not affecting DPS report contrary to the other OHKO abilities ? Are you sure about this ?

Exeatop:
The attacker is null and the damage is true, so the only relevant DPS field to be updated is shield damage (if any shield). I suppose that is a small effect on the DPS meter. Shield can be set to 0 beforehand and then there won't be any effect on the meter. Other OHKO abilities have the issue that there is an attacker field.

Exeatop:
The pillar is being deleted. Why should shield damage be on the DPS report? There are cases to be made for and against shield damage appearing on the DPS report for other OHKO abilities, but pillar removal shouldn't appear on the DPS report at all imo (as is currently the case).

sylvainpolletvillard:
OK we agree

PR #2688, opened 2025-01-11
Effect refactor extension: resetStacks, MonsterKillEffect, FireHitEffect

Exeatop:
- Added optional resetStacks method to the base Effect class for use in resurrection
- Added resetStacks method to the GrowGroundEffect class
    - May cause Big Nugget to grant gold twice. Balance decision to be decided. I'm personally ok with this.
- Expanded OnKillEffect to be used with synergy effects
- Refactored Fire and Monster stacking effects (complete with resetStacks)

Next I'm handling Sound, and then item stacks so the resurrect bugs can finally be fixed.

### DIFF
--- a/app/core/effect.ts
+++ b/app/core/effect.ts
@@ -3,15 +3,22 @@ import { PokemonEntity } from "./pokemon-entity"
 import { Item } from "../types/enum/Item"
 import { Effect as EffectEnum } from "../types/enum/Effect"
 import { Synergy, SynergyEffects } from "../types/enum/Synergy"
+import { max } from "../utils/number"
 
 type EffectOrigin = EffectEnum | Item
 
 export abstract class Effect {
   origin?: EffectOrigin
   apply: (entity: PokemonEntity, ...others: any[]) => void
-  constructor(effect: (entity: PokemonEntity, ...others: any[]) => void, origin?: EffectOrigin) {
+  resetStacks?: (entity: PokemonEntity) => void
+  constructor(
+    effect: (entity: PokemonEntity, ...others: any[]) => void,
+    origin?: EffectOrigin,
+    resetStacks?: (entity: PokemonEntity) => void
+  ) {
     this.apply = effect
     this.origin = origin
+    this.resetStacks = resetStacks
   }
 }
 
@@ -28,9 +35,10 @@ export class PeriodicEffect extends Effect {
   constructor(
     effect: (entity: PokemonEntity) => void,
     intervalMs: number,
-    origin?: EffectOrigin
+    origin?: EffectOrigin,
+    resetStacks?: (entity: PokemonEntity) => void
   ) {
-    super(effect, origin)
+    super(effect, origin, resetStacks)
     this.intervalMs = intervalMs
     this.timer = intervalMs
     this.count = 0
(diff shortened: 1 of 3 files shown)


app/core/effect.ts
```diff
@@ -74,9 +93,80 @@ export class GrowGroundEffect extends PeriodicEffect {
...
+      (pokemon, target, board) => {
+        pokemon.addAttack(synergyLevel, pokemon, 0, false)
+        if (chance(burnChance, pokemon)) {
+          target.status.triggerBurn(2000, target, pokemon)
+        }
+        this.count += 1
+      },
+      effect,
+      (pokemon) => {
+        const removalAmount = -this.count * synergyLevel
```
sylvainpolletvillard, line 166: i don't think unit should lose their attack buffs when losing fire synergy (by stealing fire stone i suppose, being the only way to lose fire synergy during a fight)

app/core/effect.ts
```diff
@@ -74,9 +93,80 @@ export class GrowGroundEffect extends PeriodicEffect {
...
+        const lifeBoost = hpGain * target.hp
+        pokemon.addAttack(attackBoost, pokemon, 0, false)
+        pokemon.addAbilityPower(apBoost, pokemon, 0, false)
+        pokemon.addMaxHP(lifeBoost, pokemon, 0, false)
+        this.hpBoosted += lifeBoost
+        this.count += 1
+      },
+      effect,
+      (pokemon) => {
+        pokemon.addAttack(-this.count * attackBoost, pokemon, 0, false)
```
sylvainpolletvillard, line 127: i don't think unit should lose their  buffs when losing synergy

sylvainpolletvillard:
Removing a synergy effect doesn't mean removing the stat buffs that have been previously given by this effect

Exeatop:
> Removing a synergy effect doesn't mean removing the stat buffs that have been previously given by this effect

As stated in my description, this method is for use in resurrection not in the loss of synergy. It's to be used for synergies that have stacking effects which are intended to be reset on resurrection. 

> - Added optional resetStacks method to the base Effect class for use in resurrection

app/core/effect.ts
```diff
@@ -3,15 +3,23 @@ import { PokemonEntity } from "./pokemon-entity"
...
+import { max } from "../utils/number"
+import { chance } from "../utils/random"
 
 type EffectOrigin = EffectEnum | Item
 
 export abstract class Effect {
   origin?: EffectOrigin
   apply: (entity: PokemonEntity, ...others: any[]) => void
-  constructor(effect: (entity: PokemonEntity, ...others: any[]) => void, origin?: EffectOrigin) {
+  resetStacks?: (entity: PokemonEntity) => void
```
sylvainpolletvillard, line 14: suggestion: subclass StackableEffect with a `stacks` or `count` prop and the resetStacks method

Exeatop: How would the stacks property be incremented? Having the resetStacks and apply methods on the same class allows them to share the same counter. If you're saying they should extend a `StackableEffect` class instead, that doesn't work because they need to inherit from the classes based on their trigger condition and javascript doesn't support multiple inheritance directly.

Exeatop:
I'm going to reopen this PR once https://github.com/keldaanCommunity/pokemonAutoChess/pull/2693 is sorted, because this needs to be redone with those changes.

PR #2691, opened 2025-01-12
Update translations (ZH - Latest)

Benzene114:
#2680 ,#2685 & #2686


Benzene114:
commit discarded due to massive conflict. The first Fink contributor would likely face this issue again

PR #2693, opened 2025-01-12
GrowGround refactor memory improvement

Exeatop:
apply method is defined in the class itself instead of the constructor, so that multiple instances of the class share the reference to the function in memory
https://discord.com/channels/737230355039387749/1317095176333824041/1328133247389007932

### DIFF
--- a/app/core/effect.ts
+++ b/app/core/effect.ts
@@ -8,9 +8,11 @@ type EffectOrigin = EffectEnum | Item
 
 export abstract class Effect {
   origin?: EffectOrigin
-  apply: (entity: PokemonEntity, ...others: any[]) => void
-  constructor(effect: (entity: PokemonEntity, ...others: any[]) => void, origin?: EffectOrigin) {
-    this.apply = effect
+  apply(entity: PokemonEntity, ...others: any[]) {}
+  constructor(effect?: (entity: PokemonEntity, ...others: any[]) => void, origin?: EffectOrigin) {
+    if (effect) {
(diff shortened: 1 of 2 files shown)


app/core/pokemon-entity.ts
```diff
@@ -122,7 +122,7 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
   isClone = false
   refToBoardPokemon: IPokemon
   commands = new Array<SimulationCommand>()
-  effectsSet = new Set<EffectClass>()
+  effectsSet = new Set<DynamicEffect>()
```
sylvainpolletvillard, line 125: not sure what you call DynamicEffect. All effects are dynamic, since they are triggered on a certain condition ?

app/core/effect.ts
```diff
@@ -20,17 +20,35 @@ export class OnItemGainedEffect extends Effect {}
...
+  apply: (entity: PokemonEntity, target: PokemonEntity, board: Board) => void
+  constructor(
+    effect: (entity: PokemonEntity, target: PokemonEntity, board: Board) => void
+  ) {
+    super(effect)
+    this.apply = effect
+  }
+}
+
+export abstract class DynamicEffect {
```
sylvainpolletvillard, line 34: i dont understand the reasoning behind having another abstract class, Effect & DynamicEffect. If its just about the constructor signature to have the apply logic defined in a prototype method vs an own property passed in constructor, then just make the argument in the constructor optional

PR #2701, opened 2025-01-15
Fixed typos

nikhil-dronamraju:
Fixed some typos where it says ennemy/ennemies (should be enemy/enemies)
- abilities.ts
- Talonflame in translation.json


Benzene114:
I'm pretty sure "POISONNED" is an established keyword as seen in Status.ts; to change that would require a change on this file as well every translation files since all-caps keywords are kept as-is

sylvainpolletvillard:
Uppercase words need to be kept as is

nikhil-dronamraju:
Reverted changes, lmk if there's any other issues

PR #2708, opened 2025-01-17
Effect refactor extension: resetStacks, MonsterKillEffect, FireHitEffect

Exeatop:
- Added optional resetStacks method to the base Effect class for use in resurrection
- Added resetStacks method to the GrowGroundEffect class
    - May cause Big Nugget to grant gold twice. Balance decision to be decided. I'm personally ok with this.
- Expanded OnKillEffect to be used with synergy effects
- Refactored Fire and Monster stacking effects (complete with resetStacks)


Exeatop:
> suggestion: subclass StackableEffect with a `stacks` or `count` prop and the resetStacks method

Having the resetStacks and apply methods on the same class allows them to share the same counter. If this means they should extend a StackableEffect class instead, that doesn't work because they need to inherit from the classes based on their trigger condition and javascript doesn't support multiple inheritance directly.

PR #2717, opened 2025-01-19
Effect refactor: sound ability effect

Exeatop:
Also refactored Chimecho and Clauncher passives.

This refactor was done with a side objective of minimizing the number of times the board is searched so that it is only done when necessary. This is the reason Chimecho's passive was given an effect enum.


app/core/effect.ts
```diff
@@ -152,3 +154,83 @@ export class FireHitEffect extends OnHitEffect {
     this.count = 0
   }
 }
+
+export class OnAbilityEffect extends Effect {
```
sylvainpolletvillard, line 158: suggestion naming: OnCastEffect or OnAbilityCastEffect

app/core/effect.ts
```diff
@@ -152,3 +154,83 @@ export class FireHitEffect extends OnHitEffect {
...
+        ally.addAttack(attackBoost * scale, pokemon, 0, false)
+        ally.addAttackSpeed(attackSpeedBoost * scale, pokemon, 0, false)
+        ally.addPP(manaBoost * scale, pokemon, 0, false)
+        ally.count.soundCryCount += scale
+      }
+    })
+  }
+}
+
+export class ResetSoundCry extends Effect {
```
sylvainpolletvillard, line 216: why is this a separate effect from SoundCryEffect instead of the resetStacks method of SoundCryEffect ?

app/core/effect.ts
```diff
@@ -152,3 +154,83 @@ export class FireHitEffect extends OnHitEffect {
...
+    }
+  }
+
+  apply(pokemon, state, board, target, crit) {
+    const attackBoost = [2, 1, 1][this.synergyLevel] ?? 0
+    const attackSpeedBoost = [0, 5, 5][this.synergyLevel] ?? 0
+    const manaBoost = [0, 0, 3][this.synergyLevel] ?? 0
+
+    const chimecho =
+      pokemon.effects.has(EffectEnum.CHIMECHO_PASSIVE) &&
```
sylvainpolletvillard, line 196: i dont understand why this new effect is needed. If you find a Chimecho adjacent and you're a sound type, then we already know you have this effect ?

app/core/effect.ts
Exeatop: The application conditions answer this question. 
```
 if (types.has(Synergy.SOUND)) {
          pokemon.effects.add(effect)
          pokemon.effectsSet.add(new SoundCryEffect(effect))
        }
        pokemon.effectsSet.add(new ResetSoundCry(effect))
```
Only sound pokemon should have the SoundCryEffect, but all pokemon on the team need to be able to reset their stacks.

app/core/effect.ts
Exeatop: The order is the opposite. If you have the sound type and the Chimecho effect is in play, then look for an adjacent Chimecho. It trades having the extra effect when a Chimecho is present for searching through the board when Chimecho is not present.

app/core/effect.ts
sylvainpolletvillard: why is that ? you add the SoundCryEffect when you steal a Metronome ?

app/core/effect.ts
sylvainpolletvillard: but you're only looking up the 8 adjacent cells, not the entire board ? That seems to be a more complicated solution for a micro optimization

sylvainpolletvillard: also since this CHIMECHO_PASSIVE effect is applied in applyPostEffects, that means it doesn't work if the sound pokemon has been added after the start, with spawn, transform or metronome stolen

app/core/effect.ts
Exeatop: I'm not sure exactly what you're asking.

```
if (pokemon.types.has(Synergy.SOUND)) {
  soundBoost(pokemon, board)
  if (pokemon.passive === Passive.MEGA_LAUNCHER) {
    soundBoost(pokemon, board)
    soundBoost(pokemon, board)
  }
}
```
The SoundCryEffect is equivalent to the current implementation's `soundBoost` function, which is only called for pokemon with the sound synergy. That's why there's the condition that only sound pokemon get `SoundCryEffect`.

However `resetStacks` can't be a method on `SoundCryEffect`, because its `apply` method affects every pokemon on the team. When it comes time to reset all of the synergy stacks during resurrection, every pokemon on the team is going to have stacks that should be reset - regardless of type. That's why the `resetStacks` method has to be on a different class in this case.

app/core/effect.ts
Exeatop: the CHIMECHO_PASSIVE effect is applied in applyEffect, so it applies to spawned pokemon. Pokemon that steal a metronome wouldn't get it though. I suppose there's no need to limit its application to sound pokemon - the entire team could gain that effect and then it would be applied on metronome gain. Transform doesn't change synergies anyways so I don't think it's relevant in this case.

 Without an effect, each sound pokemon will have to check 8 cells every time they cast an ability - which happens to be frequently for sound types. That's an improvement over the current implementation, but it's still a lot of excess checking. I don't have a problem with scrapping this effect enum if you think the tradeoff isn't worthwhile.

app/core/effect.ts
sylvainpolletvillard: > When it comes time to reset all of the synergy stacks during resurrection, every pokemon on the team is going to have stacks that should be reset - regardless of type.

What are you talking about ? In what scenario a resurrection should reset stacks of other mons than the one resurrected

app/core/effect.ts
sylvainpolletvillard: We have abilities with much more cells checked. This looks like premature optimization and giving a "CHIMECHO_PASSIVE" to the entire team feels very confusing for me. I would suggest simply removing this effect, the rest should work just fine

app/core/effect.ts
sylvainpolletvillard: Okay I finally understood after struggling for way too long. You're talking specifically about **not sound** units that could have benefit from sound cries stat buffs, and those buffs needs to be cancelled after resurection. Now i understand better your message on [Discord](https://discord.com/channels/737230355039387749/1317095176333824041/1328073667183382619)

Not gonna lie, since you introduced this "resetStacks" method in https://github.com/keldaanCommunity/pokemonAutoChess/pull/2688 (that still isnt merged by the way, so this resetStacks in this PR does nothing), I'm having a really hard time following the logic. I think this workaround with another effect that is only useful for managing resurection usecase is a clear indication that this is not the right approach.

Exeatop: The branch in https://github.com/keldaanCommunity/pokemonAutoChess/pull/2688 was annoying me with back merging so those changes were moved to a new branch which was merged in https://github.com/keldaanCommunity/pokemonAutoChess/pull/2708. It won't do anything until the resurrect method is updated anyways.

> a clear indication that this is not the right approach
I see your point here. My handling of resetting item stacks was much more elegant. It might be easier for you to understand my intent if you see how I was planning to change resurrect. You can see that from the pokemon-entity.ts file linked here: https://github.com/keldaanCommunity/pokemonAutoChess/compare/master...Exeatop:pokemonAutoChess:resurrect-refactor#diff-20b1bbbbdf5f9f824bfcff0ae87fd8f6e8ec93523b54162a9f394920107fcfae

Note that the changes there aren't ready to be merged for a few reasons, but it does outline the intent.

It's possible to add the stack resetting logic directly to the resurrect class and remove the need for a resetStacks method, but it's a bit messier for effects other than sound. Maybe it's best if only the sound reset logic is in the resurrect method since the count variable is directly attached to the pokemon anyways. What do you think about that?

sylvainpolletvillard: Agree, centralizing all the effects of resurrect in one function will also help having a clear vision of what resurrection should reset or not. I never thought of sound buffs initially for example.

Exeatop: I'll handle this when I update the resurrect method.

sylvainpolletvillard:
Good to merge for me 👍

PR #2726, opened 2025-01-21
remove flora from torterra

sylvainpolletvillard:
Remove flora type of turtwig line in an effort to reduce overlap between grass and flora

There are several other ways to get max grass max flora so this won't have much impact except lowering average flora when playing grass, acting as an indirect nerf.

Turtwig and Torterra do not have flower in their design too


ChorizardAC:
agree

jlgrimes:
nah it's fine i think

PR #2730, opened 2025-01-22
Refactor resurrect

Exeatop:
https://discord.com/channels/737230355039387749/1317095176333824041/1320421361063034954
> Resurrection remove statuses with some exceptions, i think of enraged, tapu fields, and all the things implemented as statuses but not actual statuses in the game.
Resurrection also reset item and synergy stacks, and their associated stat buffs. This has been decided as a balance measure
the rest shall remain unaffected

### DIFF
--- a/app/core/effect.ts
+++ b/app/core/effect.ts
@@ -57,15 +57,6 @@ export class MonsterKillEffect extends OnKillEffect {
     this.count += 1
   }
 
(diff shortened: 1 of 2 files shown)


sylvainpolletvillard:
Great, this is much better ! Easy to read and to adjust if we feel the need for it

PR #2741, opened 2025-01-26
add comfey item gained and removal effects

Exeatop:
~~Comfey's stats can vary wildly based on the effects in play, as well as its position on the board (for example, in light cell or next to Stonjourner). These aren't necessarily repeatable by simply spawning in a new Comfey, so the only way to be able to remove these stats again (and apply them when stolen) is to record them somewhere. I chose to put these in an Effect subclass but I'm open to hearing about better solutions in case I missed something.~~

Comfey now only grants base stats to its holder, in addition to having item removal effects.


sylvainpolletvillard:
> Comfey's stats can vary wildly based on the effects in play
Which ones ? It cannot hold items

Exeatop:
Comfey's stats change based on board location from Light, Normal, and Stonjourner. I don't think there are currently any others that could impact it

sylvainpolletvillard:
Those stats do not need to be carried, you can just use base stats

Exeatop:
> Those stats do not need to be carried, you can just use base stats

Are you saying you don't want any synergy effects to be carried? They're currently being carried so that would be a nerf. It would make the logic much easier though.

sylvainpolletvillard:
I don't know what synergy effects you're talking about, just comfey stats as a held item are base stats, the effects you mentionned are not relevant

Exeatop:
Ice special defense and amorphous attack speed are also being added. I'm in the middle of converting it to base stats so it should be ok once it's updated.

PR #2752, opened 2025-01-29
Update translations (ZH - Latest)

Benzene114:
#2744 ,#2748 &#2749


Benzene114:
plus #2741 .
The move rename is simply to make the hp move names somewhat consistent (in that every move name's Chinese part is 2 chara), 黄（圣）菇 is one of Uxie's more common unofficial Chinese name

keldaan-ag:
you're the best

PR #2766, opened 2025-02-02
handle divergent evolutions info in pokemon class

sylvainpolletvillard:
- added `evolutions` field in Pokemon class for pokemons with divergent evolutions
- Divergent evolutions are now shown in Wiki > Pokemons
- Eeveelutions now are properly considered as evolutions of Eevee. That means Eevee can hold eviolite, and Eeveelutions do not count several times for the same synergy when played together.


keldaan-ag:
👍

PR #2786, opened 2025-02-05
Hotfix pokemon with divergent evolutions don't shimmer in shop + dynamically update divergent evolution portraits

Exeatop:
https://discord.com/channels/737230355039387749/1336513648905945141

~~It might be possible to display the proper evolution for divergent evolutions. I'll look into that.~~ Done

### DIFF
--- a/app/core/evolution-rules.ts
+++ b/app/core/evolution-rules.ts
@@ -1,6 +1,7 @@
 import Player from "../models/colyseus-models/player"
 import { Pokemon, PokemonClasses } from "../models/colyseus-models/pokemon"
 import PokemonFactory from "../models/pokemon-factory"
+import { IPlayer } from "../types"
 import { EvolutionTime } from "../types/Config"
 import { Ability } from "../types/enum/Ability"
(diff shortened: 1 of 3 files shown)


sylvainpolletvillard:
Beware pokemon.evolution is not always defined, for divergent evolutions you want to look at pokemon evolutions array

This could be the opportunity to fix evolution portraits to work with divergent evolutions too

Exeatop:
> Beware pokemon.evolution is not always defined, for divergent evolutions you want to look at pokemon evolutions array

Ah yeah I handled this in my fix for dynamic portraits but it looks like I forgot to update it here. Should I just submit that in one PR all at once?

Exeatop:
One issue remains. Pokemon whose evolution conditions can change during a round (like poliwhirl) don't have their evolution icon updated automatically; the shop has to be updated for the icon to update. For such pokemon, perhaps we should use `onChangePosition` to update the shop.

sylvainpolletvillard:
It's already much better, good job !

PR #2790, opened 2025-02-05
Update DE translations

Talessia:
Added missing ability translations
Added missing Item descriptions
Added missing weather descriptions


Exeatop:
Repaired file is here:
https://github.com/Exeatop/pokemonAutoChess/blob/de-translation/app/public/dist/client/locales/de/translation.json

PR #2804, opened 2025-02-08
Prevent out of bounds drag drop exploit

WhatIsACore:
Add checks to make sure drag drop target is within the board. Prevents abuses where a pokemon can start on the opponent's side of the board, or a pokemon can be placed out of bounds (making them untargetable, but still able to attack opponents)

https://discord.com/channels/737230355039387749/1337590750451470438


keldaan-ag:
use BOARD_HEIGHT & BOARD_WIDTH from  ../types/Config

PR #2807, opened 2025-02-08
Add antialiasing support, refactor preferences a little bit

whisperdoll:
- Add antialiasing preference
  - Change styles accordingly
- Add `PokemonPortrait` component to DRY up some images
- Refactor preferences a bit
  - Add subscriptions and React hooks so components can update in response to preferences changing
  - Freeze object and remove export to prevent accidental writes to preferences without going through `savePreferences` (and notifying subscribers, etc)
  - Move volume update logic to `playMusic` so options modals etc don't have to concerns themselves with it
  - Remove extrenuous instances of `loadPreferences` and don't export it since they're loaded initially anyway
- Clean up keybind modal a little bit
  - Rename `keyRemapped` -> `currentlyRemapping` for clarity
  - Memoize

lmk if it's too big and ill break it up. or if you just generally dont fw the concept

### DIFF
--- a/app/public/src/style/toastify.css
+++ b/app/public/src/style/toastify.css
@@ -19,17 +19,4 @@
 .toast-new-pokemon img {
   width: 40px;
   height: 40px;
-  image-rendering: pixelated;
-}
-
-.toast-emote {
-  right: 85px !important;
-  border: none !important;
-  background-color: transparent !important;
-}
-
-.toast-emote img {
-  width: 80px;
-  height: 80px;
-  image-rendering: pixelated;
 }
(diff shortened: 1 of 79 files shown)


app/public/src/style/toastify.css
```diff
@@ -19,17 +19,4 @@
 .toast-new-pokemon img {
   width: 40px;
   height: 40px;
-  image-rendering: pixelated;
-}
-
-.toast-emote {
-  right: 85px !important;
```
sylvainpolletvillard, line 26: Why are those 2 selectors deleted ?

sylvainpolletvillard:
The refactor looks good. The antialias looks bad but I guess it will be up to players to decide on that 😅

app/public/src/style/toastify.css
whisperdoll: theyre unused as far as i can tell

sylvainpolletvillard: ok yeah it seems the real used class is "game-emote-bubble"

PR #2831, opened 2025-02-15
Fix Nuzzle and Accelerock targeting

Exeatop:
https://discord.com/channels/737230355039387749/1339870329912164362

### DIFF
--- a/app/core/abilities/abilities.ts
+++ b/app/core/abilities/abilities.ts
@@ -1942,15 +1942,15 @@ export class AccelerockStrategy extends AbilityStrategy {
       pokemon.moveTo(destination.x, destination.y, board)
       pokemon.targetX = destination.target.positionX
       pokemon.targetY = destination.target.positionY
-      target.handleSpecialDamage(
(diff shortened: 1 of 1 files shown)


app/core/abilities/abilities.ts
```diff
@@ -1934,31 +1934,22 @@ export class AccelerockStrategy extends AbilityStrategy {
...
   ) {
-    super.process(pokemon, state, board, target, crit)
-    const farthestCoordinate =
+    const destination =
       board.getFarthestTargetCoordinateAvailablePlace(pokemon)
-
-    if (farthestCoordinate) {
-      pokemon.moveTo(farthestCoordinate.x, farthestCoordinate.y, board)
-      const targetCoordinates = state.getNearestTargetAtRangeCoordinates(
+    target = destination?.target ?? target
```
sylvainpolletvillard, line 1939: if no destination is found, the fallback to target wont be useful because of the if(destination) condition

maybe the target.handleSpecialDamage() should be outside of that condition ?

PR #2832, opened 2025-02-16
Balance-16-02-25

sylvainpolletvillard:
# Changes to Pokemon & Abilities

- Leafeon is no longer Flora
- Vaporeon is no longer Amorphous
- Flareon is no longer Light
- Sylveon is no longer Sound
- Buff Vulpix line andd Alolan Vulpix line: PP 85 → 80 : Aurora Veil Rune protect duration 0.5 → 1 second
- Nerf Teatime (Polteageist): healing 20/40/80 → 15/30/60
- Buff Houndour line: PP 125 → 110
- Buff Sneasel Attack 9 → 10 and Weavile: Attack 22 → 26
- Buff Topsy Turvy (Inkay): also reverse AP buffs
- Buff Pawniard Attack 14 → 17 and Bisharp Attack 26 → 30
- Nerf Inteleon: Snipe shot damage 40/80/160 → 40/80/120
- Nerf Stantler: HP 200 → 180, Def/Spedef: 5 → 3
- Nerf Miltank: HP 250 → 200
- Nerf Spinda: HP 250 → 200
- Nerf Burmy line: Burmy now has various defense stats based on its cloak: Plant Cloak: def 3/3/3 → 1/1/1, spedef 3/5/3 → 1/2/3, Sand Cloak def 3/3/3 → 2/2/2, spedef 3/5/3 → 2/3/3
- Buff Melmetal: increase Attack speed/Attack ratio: 10% attack speed = ~~3~~ 5 attack

# Changes to Items

- Nerf Pokerus: effect applies every ~~2~~ 2.5 seconds
- Buff Swift Wing: 25 → 30% attack speed
- Minor nerf to Repeat Ball: legendaries are found afer ~~140~~ 150 rolls

PR #2845, opened 2025-02-19
Update Translations (ZH - Latest + Misc - Missing Recent Balances)

Benzene114:
#2841 ,#2842 ,& #2844 
A few balance changes in #2832 wasn't shown in translation files (Melmetal Passive buff and Snipe Shot nerf), hence the edit to every file that has these translated


sylvainpolletvillard:
Good catch, thank you !

PR #2861, opened 2025-02-23
New speed stat

sylvainpolletvillard:
- Attack speed, move speed and ability cooldowns have all been merged into one statistic: Speed. 1 Speed = 1% attack speed in previous patch. Pokemons now have a different base speed as part of their stats, taken from their base game stats. Many attack and max PP stats have been adjusted to compensate for the difference of base speed.

- Added collect-speed-stat script that copies the speed stat from PokeAPI
- Added add-speed-stat script that rewrites pokemon classes fields to add the speed and adjust attack/maxPP

- Changed Wild to work with the new speed stat: Wild now gives Speed at levels 2-4 and Attack at levels 6-8 instead of attack and movement speed at all levels

- Weather Windy and Snow now impacts speed instead of movement speed

- Changed pokemon tooltips to show speed stat. AP and crit stats removed from board pokemon tooltip


sylvainpolletvillard:
image
 new board tooltip

PR #2863, opened 2025-02-24
Balance-24-02-15

sylvainpolletvillard:
- Changed Heavy Slam (Aggron): Shield 10/20/40 → 15/30/60, remove AP scaling
- Buff Concrete Pillar (Conkeldurr) HP 300 → 400

HeavySlam is actually bugged this patch, its already 15/30/60 shield, so its a straight up nerf


keldaan-ag:
👍

PR #2864, opened 2025-02-24
Standardize EN terminology for KO

whisperdoll:
In the future it would be nice to give it a tooltip in `addIconsToDescription` too


sylvainpolletvillard:
"KOing" is that a real word ?

sylvainpolletvillard:
Okay for replacing knocking down by knocking out 👍 

It's mostly the KO'd and KOing forms that look weird to me

Exeatop:
The usage of the official titles is either "KO" or "faints." "If the target faints, X happens."

PR #2869, opened 2025-02-25
Fix evolutions not shown as on team in the Synergy sidebar - Better implementation of #2868

Exeatop:
Improves https://github.com/keldaanCommunity/pokemonAutoChess/pull/2868
Middle forms gaining a type wasn't considered
https://discord.com/channels/737230355039387749/1343754272575062156


sylvainpolletvillard:
This is not better, its using out of scope variable to hold state. 

A better implementation would be to add "checked" property for PokemonPortrait, passed as prop

sylvainpolletvillard:
The new prop needs to be optional, PokemonPortait is used in many places

Exeatop:
> The new prop needs to be optional, PokemonPortait is used in many places

That's a different function with the same name. This one isn't exported

PR #2885, opened 2025-03-04
adjust component rule at stage 19

sylvainpolletvillard:
Instead of cancelling stage 17 carousel encounter in case of odd component count, i compute the number of components owned by the player before stage 19, and give full items if even number. This is more safe to avoid useless remaining component in more scenarios like unown-I


keldaan-ag:
👍

PR #2898, opened 2025-03-08
show tms in wiki abilities

sylvainpolletvillard:
show tms in wiki abilities

also nerf rotom drone pp

chrome_k7JrJYbaDf


keldaan-ag:
👍

PR #2900, opened 2025-03-08
Buff paralysis

sylvainpolletvillard:
buff paralysis to reduce speed by 50 instead of 40, as a compensation for increased base speed for some mons that are countered by paralysis

also fixed a bug that gave back too much speed when paralysis was removed on a mon that already had less than 40 speed


keldaan-ag:
👍

PR #2907, opened 2025-03-10
add regirock encounter

sylvainpolletvillard:
Went with additional HP for all players. Feel free to propose something else

Also change it so every encounter can only be found once per game


keldaan-ag:
👍

PR #2912, opened 2025-03-13
Balance-13-03-25

sylvainpolletvillard:
- Revert Inteleon nerf
- Change Jirachi buff: PP 100 → 80. Reduce ability cooldown to 200ms at 50 speed.
- Electric synergy levels are now 3/5/7 ; change Electric 7: Triple attacks also deal ~~20 special damage to enemies adjacent to the target~~ 10 special damage to **target and enemies adjacent to it**


keldaan-ag:
👍

PR #2943, opened 2025-03-21
Fix-refreshShopUI

sylvainpolletvillard:
fix https://discord.com/channels/737230355039387749/1352254438945062952


Exeatop:
Ah that's what I was supposed to do. Makes sense. My bad. Looks good, I tested it 👍

PR #2946, opened 2025-03-21
Hotfix turn 19 giving items when it's expected to give components

Exeatop:
https://discord.com/channels/737230355039387749/1352371040642404402


sylvainpolletvillard:
good catch. Not sure if considered a bug of ArraySchema

PR #2992, opened 2025-03-28
Prevent Beat Up from summoning a Houndour with 0hp

Exeatop:
https://discord.com/channels/737230355039387749/1355228940905742396

### DIFF
--- a/app/core/abilities/abilities.ts
+++ b/app/core/abilities/abilities.ts
@@ -258,8 +258,8 @@ export class BeatUpStrategy extends AbilityStrategy {
         true
       )
       const scale = (1 + pokemon.ap / 100) * (crit ? pokemon.critPower : 1)
-      entity.hp = Math.round(houndour.hp * scale)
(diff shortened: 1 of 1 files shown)


app/core/abilities/abilities.ts
```diff
@@ -258,8 +258,8 @@ export class BeatUpStrategy extends AbilityStrategy {
         true
       )
       const scale = (1 + pokemon.ap / 100) * (crit ? pokemon.critPower : 1)
-      entity.hp = Math.round(houndour.hp * scale)
-      entity.life = Math.round(houndour.hp * scale)
+      entity.hp = min(1)(Math.round(houndour.hp * scale))
```
ahorner, line 261:
```suggestion
      entity.hp = Math.min(1, Math.round(houndour.hp * scale))
```

Exeatop: This is wrong. `min` is a custom curried function equivalent to `Math.max`

PR #3011, opened 2025-04-06
Starly to common regional

sylvainpolletvillard:
Moved Starly line from hatch to common to give an incentive for normal/flying portals stage 0

Another hatch flying would have to be added later (3 other 3 stage flying lines remaining)


app/models/colyseus-models/pokemon.ts
```diff
@@ -10353,53 +10353,54 @@ export class Serperior extends Pokemon {
 
 export class Starly extends Pokemon {
   types = new SetSchema<Synergy>([Synergy.NORMAL, Synergy.FLYING])
-  rarity = Rarity.HATCH
+  rarity = Rarity.COMMON
   stars = 1
   evolution = Pkm.STARAVIA
   evolutionRule = new HatchEvolutionRule(EvolutionTime.EVOLVE_HATCH)
```
Benzene114, line 10359: Is this line supposed to be kept?

PR #3012, opened 2025-04-06
regional pokemons sold in a region other than their original region are not added back to the pool

sylvainpolletvillard:
fix https://discord.com/channels/737230355039387749/1358276980067270836


keldaan-ag:
👍

PR #3060, opened 2025-04-22
Fix Unown X bug

Phoenix-Starlight:
When X-Ray was already equipped and Unown X casted, it would add the speed buff despite being already applied.

### DIFF
--- a/app/core/pokemon-entity.ts
+++ b/app/core/pokemon-entity.ts
@@ -620,8 +620,9 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
     ) {
       return
     }
-    this.items.add(item)
-    this.simulation.applyItemEffect(this, item)
+    if (this.items.add(item)) {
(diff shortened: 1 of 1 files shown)


sylvainpolletvillard:
Set#add doesn't return a boolean, it returns the set itself

app/core/pokemon-entity.ts
```diff
@@ -620,8 +620,9 @@ export class PokemonEntity extends Schema implements IPokemonEntity {
     ) {
       return
     }
-    this.items.add(item)
-    this.simulation.applyItemEffect(this, item)
+    if (this.items.add(item)) {
```
sylvainpolletvillard, line 623:
```suggestion
    if (this.items.has(item) === false) {
        this.items.add(item)        
```

sylvainpolletvillard (changes requested):

PR #3061, opened 2025-04-23
fix devolution of archaludon

sylvainpolletvillard:
fix archaludon not devolving when removing an artificial item that is not a synergy item

it's still a very bug-prone mechanic, i dont like it


sylvainpolletvillard:
as you wish

PR #3084, opened 2025-04-28
redesigned HP / PP bars

WhatIsACore:
Currently, client-side pokemon HP / PP bars borrow their appearance and animations from the `progress-bar.tsx` component. This redesigns the HP / PP bars to have snappier animations, visual segmentation of every 20 HP, as well as combines them under a complete LifeBar class that manages both. Shields appear as white and add onto the HP bar with better visual logic now as well. The redundant `bar.ts` and `powerbar.ts` classes have been removed.

https://github.com/user-attachments/assets/fc3f637b-8c1a-4013-ade9-e77a785df9c6

The code pattern is slightly changed: instead of calling `pokemon.lifebar.setAmount` and `pokemon.powerbar.setAmount` methods separately, if for some reason there's a need to update the visual appearance of client-side health and pp bars it's changed to:

`pokemon.lifebar.setLife(x)`
`pokemon.lifebar.setMaxLife(x)`
`pokemon.lifebar.setPP(x)`
`pokemon.lifebar.setMaxPP(x)`

This doesn't affect any non-visual game logic and only interacts with the rest of the code at three points (where the pokemon, pokemon-avatar, and battle-manager initiate or update the healthbar components) where I've already swapped the methods out. The PP bar only appears if the pp or max pp are ever set, so avatars and pillars still have their PP bars hidden like before.


WhatIsACore:
I didn't consult with anyone before trying this so if this change isn't desired or is not welcome feel free to reject!

WhatIsACore:
Rewritten as a canvas implementation:

https://github.com/user-attachments/assets/29c9b571-0269-4b8e-bcb8-6265d89b548e

PR #3086, opened 2025-04-30
Hotifx Reaper Cloth doesn't allow abilities to crit

Exeatop:
https://discord.com/channels/737230355039387749/1367018879762698240

There is an edge case where if Reaper Cloth is removed but the holder was given a Leek it will still remove the ability to crit. I don't know how to handle that but also it might be ok to just not address as a very rare edge case.


sylvainpolletvillard:
Oopsie. Looks like reaper cloth effect code has disappeared during the v6 merge.

PR #3092, opened 2025-05-02
replace pastebin by approval system for bot managers

sylvainpolletvillard:
Because pastebin service was causing lots of issues for several months, i replaced it by an approval system

The idea is that every bot submitted is directly saved to database with an "approved" flag set to false. Then bot managers through the bot management panel can watch them and approve them. Once approved, they will be picked and visible for custom lobbies.

All the websocket transfer calls have been replaced by GET/POST/DELETE HTTP calls, more appropriate for this amount of data.

https://discord.com/channels/737230355039387749/1320052925174775868

❗ A db-command script has been added to set all existing bots approved flag to true. Optionnally, we can skip that script and let bot managers approve all the existing bots one by one


sylvainpolletvillard:
chrome_XF2Jk68DhL

PR #3093, opened 2025-05-02
Stat Application Refactor

Exeatop:
Refactor stat application methods (`PokemonEntity.add[stat]`) to route through `applyStat`.

This is needed to prevent logic duplication in my improvement project: allowing item and synergy stats to be calculated and displayed in the preparation phase. Has the added benefit of allowing individual stat application methods (see Melmetal).

### DIFF
--- a/app/core/abilities/abilities.ts
+++ b/app/core/abilities/abilities.ts
@@ -39,7 +39,7 @@ import {
   EvolutionTime
 } from "../../types/Config"
 import { Effect } from "../../types/enum/Effect"
-import { AttackType, Orientation, Team } from "../../types/enum/Game"
+import { AttackType, Orientation, Stat, Team } from "../../types/enum/Game"
 import { ArtificialItems, Berries, Item } from "../../types/enum/Item"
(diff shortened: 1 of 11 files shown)


sylvainpolletvillard:
I dont really understand the purpose of this, not sure what you have in mind for your "improvement project", but what does this applyStat function bring to the table ? Since the individual addDef/addLuck functions are still there and called

Exeatop:
The primary objective is to move the stat application logic to the `Pokemon` class, so that both `Pokemon` and `PokemonEntity` instances have access to it. In order for stat calculation to occur during the preparation phase, `Pokemon` instances need access to these methods which are otherwise unavailable to them. I imagine the remaining stats like crit chance, crit power, and shield will also exist on the `Pokemon` class in the future for the sake of display in the preparation phase. There is currently a problem with too many stats in the tooltip (hence why range isn’t on entity tooltips) but that can be addressed

sylvainpolletvillard:
You could have just kept the existing API to do that. Not sure a single method signature for all stat buffs/debuffs is appropriate

PR #3095, opened 2025-05-03
use addMaxHP instead of handleSpecialDamage for fillet away

WhatIsACore:
The fillet away ability currently works by setting Max HP to half, and then calculating lost life (if any), and then hitting itself with handleSpecialDamage using true damage. This is intended to clamp the current life, but creates a number of bugs:

- if user is enraged, or has boosts from terrain/sheer cold, it will deal extra damage to itself beyond the intended clamp
- if user has shield, it will hit the shield first, resulting in more current life than max HP
- if user is protected, it will not lose current life at all

using addMaxHP with a negative value avoids these bugs while taking advantage of the already existing clamping feature built into addMaxHP


sylvainpolletvillard:
That was my attempt to count the lost life in DPS report, but you're right, there's too many issues with this approach

PR #3096, opened 2025-05-03
zygarde rework

sylvainpolletvillard:
Rework Zygarde: Zygarde 50% range 2 → 3 ; Zygarde 100% range 1 → 2 PP 120 → 100 ; Zygarde now evolves into 100% form after getting max stacks of ground (15 seconds) ; Rework Core Enforcer: now hits in an area of effect in the shape of a Z around the target, dealing 80 special damage and inflicting Silence for 3 seconds.

https://discord.com/channels/737230355039387749/1367844993020072036


keldaan-ag:
👍

PR #3097, opened 2025-05-04
pve pokemons stage entrance

sylvainpolletvillard:
Wild pokemons in PvE round now make a stage entrance during the picking phase


sylvainpolletvillard:
https://github.com/user-attachments/assets/bde160e5-9fd4-4c54-9d4e-23a6de7e5932

PR #3105, opened 2025-05-09
Fix Steel 6+ prevents Fairy additional special damage

Exeatop:
https://discord.com/channels/737230355039387749/1370210184538099722

### DIFF
--- a/app/core/pokemon-state.ts
+++ b/app/core/pokemon-state.ts
@@ -111,6 +111,18 @@ export default abstract class PokemonState {
         specialDamage += Math.ceil(damage * additionalSpecialDamagePart)
       }
 
+      if (target.effects.has(Effect.WONDER_ROOM)) {
+        specialDamage = Math.ceil(damage * (1 + pokemon.ap / 100))
+      } else if (pokemon.attackType === AttackType.SPECIAL) {
+        specialDamage = damage
+      } else {
+        physicalDamage = damage
+      }
+
+      if (pokemon.passive === Passive.SPOT_PANDA && target.status.confusion) {
+        specialDamage += 1 * damage * (1 + pokemon.ap / 100)
+      }
+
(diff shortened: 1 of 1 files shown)


sylvainpolletvillard:
there's more to do here, same problem for Wonder Room effect and Spot Panda passive.

sylvainpolletvillard:
No still not good, this time physicalDamage is not reduced if trueDamagePart > 0, applying the damage twice

app/core/pokemon-state.ts
```diff
@@ -107,6 +107,22 @@ export default abstract class PokemonState {
         damage = 0
       }
 
+      if (additionalSpecialDamagePart > 0) {
+        specialDamage += Math.ceil(damage * additionalSpecialDamagePart)
+      }
+
+      if (target.effects.has(Effect.WONDER_ROOM)) {
```
sylvainpolletvillard, line 114:
```suggestion
      if (target.effects.has(Effect.WONDER_ROOM)) {
        attackType = AttackType.SPECIAL
        damage = Math.ceil(damage * (1 + pokemon.ap / 100))
      }
```

app/core/pokemon-state.ts
```diff
@@ -140,22 +156,6 @@ export default abstract class PokemonState {
         totalTakenDamage += takenDamage
       }
```
sylvainpolletvillard, line 158: then after the trueDamagePart computed, line 158:
```suggestion
if (attackType === AttackType.SPECIAL) {
        specialDamage = damage
      } else {
        physicalDamage = damage
      }
```

sylvainpolletvillard (changes requested):
Need to compute in this order:
1. wonder room damage buff
2. true damage part
3. spread the remaining damage in physicalDamage or specialDamage based on attack type

PR #3111, opened 2025-05-10
Fix edge case for special damage calculation

KevinLeoA:
Fix edge case where additional special damage would be overwritten if attackType is SPECIAL (fairy xurkitree/wonder room active). Changed to add the damage instead of overwriting.


sylvainpolletvillard:
You're right, thanks, good catch

PR #3115, opened 2025-05-11
add kubfu unique

sylvainpolletvillard:
https://discord.com/channels/737230355039387749/1263934124737953886

also contain a major fix for infinite recursion with electric triple attack effect


keldaan-ag:
👍

PR #3118, opened 2025-05-13
Fix Spectral Thief using incorrect stats to calculate stolen buff

KevinLeoA:
Fix Spectral Thief using spDef to calculate defense buff.

There is also the issue of it not reducing AP and speed after stealing the buff. Should this be changed or is it intended to be like this?


sylvainpolletvillard:
> There is also the issue of it not reducing AP and speed after stealing the buff. Should this be changed or is it intended to be like this?

It should be changed. The base speed is also incorrect, its always 50 when it should be specific per pokemon

PR #3127, opened 2025-05-19
Fix Power Lens reflected damage calculation

KevinLeoA:
Fix Power Lens reflected damage calculation by using specialDamage and accounting for armor break.

Previously damageAfterReduction was calculated using 'damage' which was not boosted by AP/critical/helping hand. This made damageAfterReduction lower than it should be, which made damageBlocked higher. 

It was also not checking if the holder had armor break for speDef calculation.


sylvainpolletvillard:
Thank you !

PR #3147, opened 2025-05-28
add pending game manager

sylvainpolletvillard:
move all functions related to user timeout / pending game storage to its own file

also store and compare date to know if a player can be eliminated

trying to fix https://discord.com/channels/737230355039387749/1352629301467418654

### DIFF
--- a/app/rooms/game-room.ts
+++ b/app/rooms/game-room.ts
@@ -552,7 +552,7 @@ export default class GameRoom extends Room<GameState> {
     if (pendingGame?.gameId === this.roomId) {
       // user reconnected without reconnection token (new browser/machine/session)
       clearPendingGame(this.presence, client.auth.uid)
-    } else if (pendingGame != null && isValidDate(pendingGame.reconnectionDeadline) && pendingGame.reconnectionDeadline?.getTime() > Date.now()) {
+    } else if (pendingGame != null && !pendingGame.isExpired) {
       client.leave(CloseCodes.USER_IN_ANOTHER_GAME)
     }
   }
--- a/app/core/pending-game-manager.ts
+++ b/app/core/pending-game-manager.ts
@@ -6,6 +6,7 @@ import { ALLOWED_GAME_RECONNECTION_TIME } from "../types/Config";
 interface PendingGame {
     gameId: string;
     reconnectionDeadline: Date;
+    isExpired: boolean;
 }
 
 const PENDING_GAME = "pending_game";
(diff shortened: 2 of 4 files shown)


WhatIsACore (commented):
Fix ALLOWED_GAME_RECONNECTION_TIME and make sure to clear 5 minute loading timeout if all players successfully load

app/rooms/game-room.ts
```diff
@@ -565,22 +565,23 @@ export default class GameRoom extends Room<GameState> {
...
     } catch (e) {
       if (client && client.auth && client.auth.displayName) {
-        const pendingGameId = await this.presence.hget(
-          client.auth.uid,
-          "pending_game_id"
-        )
-        if (pendingGameId == undefined && !consented) return // user has reconnected through other ways (new browser/machine/session)
-        this.presence.hdel(client.auth.uid, "pending_game_id")
+        const pendingGame = await getPendingGame(this.presence, client.auth.uid)
+        if (!pendingGame && !consented) return // user has reconnected through other ways (new browser/machine/session)
```
WhatIsACore, line 578: Race condition that might cause random or unintended behavior.

`getPendingGame` will return `null` if pending game has expired aka after 5 minutes. However, this entire section of code only runs after `this.allowReconnection` awaits for the same 5 minutes. If the first thread is faster, then `pendingGame` will always be `null`, even if a player has actually left, and the subsequent code that removes them from the game and sets their health to -99 will not run.

app/rooms/game-room.ts
```diff
@@ -236,7 +239,7 @@ export default class GameRoom extends Room<GameState> {
       () => {
         this.broadcast(Transfer.LOADING_COMPLETE)
         this.state.players.forEach((player) => {
-          this.presence.hdel(player.id, "pending_game_id")
+          clearPendingGame(this.presence, player.id)
```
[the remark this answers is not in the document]
sylvainpolletvillard: the absence of pending game doesnt prevent a player to reconnect, but they wont have the prompt to reconnect and they wont be eliminated

app/rooms/game-room.ts
sylvainpolletvillard: but you're right, it will go through the first if branch line 578 even if not reconnected. I'm changing getPendingGame to not return null if reconnection deadline is expired

WhatIsACore (approved):
It looks good to me!

Not required but in the future we could consider this change too:

```
const pendingGame = await getPendingGame(...)
if (
    pendingGame != null &&
    isValidDate(pendingGame.reconnectionDeadline) && 
    pendingGame.reconnectionDeadline?.getTime() > Date.now()
) ... do stuff
``` 

is used multiple times. We could move that code into `getPendingGame` and return a `.isActive` boolean. Shortening the expression to:

```
const pendingGame = await getPendingGame(...)
if (pendingGame?.isActive) ... do stuff
```

PR #3148, opened 2025-05-30
PAC 6.2.0 Spanish translations

antvinper:
Every key in the English translation file now has its corresponding value in the Spanish json


sylvainpolletvillard:
Muchas gracias !

PR #3165, opened 2025-06-06
add targetEntityId to target lock

sylvainpolletvillard:
- add targetEntityId to PokemonEntity
- Targeting and aggro changes: Units now aim to continue targeting the same enemy if it moves, unless it moves out of their attack range. If the enemy moves out of range, they will pick another target.

https://discord.com/channels/737230355039387749/1376329097801633913/1377277468141944904


app/core/attacking-state.ts
```diff
@@ -93,6 +97,7 @@ export default class AttackingState extends PokemonState {
         pokemon.count.attackCount++
         pokemon.targetX = targetCoordinate.x
         pokemon.targetY = targetCoordinate.y
+        pokemon.targetEntityId = target.id
```
WhatIsACore, line 100: 1. Can use the new `pokemon.setTarget(target)`
2. If pokemon is confused, targetCoordinates will not match target.id

WhatIsACore (commented):
After reading this code, I also have a theory that the confusion effect is bugged - visually, the pokemon appears to swing in different directions but its actual target does not change. I may be wrong

WhatIsACore:
I also see three other potential bugs with this section of code but that existed even before this PR:

1. if the attacking pokemon has been moved (by aquatic wave, palkia, etc.) but its previous target has not moved, it will still attack the original target even if it's no longer within range, since a new range check does not occur except when retargeting. Can fix by adding an additional range check to line 76.

2. `getTargetCoordinatesWhenConfused` can return targets that are out of range as well, since the only check it does to filter valid targets is `isTargettableBy` which does not make a range check. Of course this is meaningless because I think confusion isn't working anyway but if we do fix confusion make sure we fix this too.

3. Just like Basic Attacks, abilities also ignore confusion.

sylvainpolletvillard:
I removed `targetCoordinate` to focus exclusively on entity targeted, that ended up simplifying the code a lot

WhatIsACore:
> if it has been moved, it should be in moving state and quit attacking state, which resets targeting

Good point, I didn't consider this. I noticed spacial rend does not cause movingState, and king's shield, poison jab, and u-turn only trigger moving state on the user but not the swapped target. since this is a different issue I opened https://github.com/keldaanCommunity/pokemonAutoChess/pull/3168

PR #3166, opened 2025-06-06
summer break aquatic 8 + surfing pikachu

keldaan-ag:


app/public/dist/client/locales/en/translation.json
```diff
@@ -3013,7 +3020,8 @@
 		"PICNICKER": "Make a picnic with 9 Pokémon",
 		"STARGAZER": "Get Solgaleo or Lunala",
 		"BLOODY": "Get Bloodmoon Ursaluna",
-		"ETERNAL": "Meet Arceus at stage 40"
+		"ETERNAL": "Meet Arceus at stage 40",
+		"SURFER": "Go on a summer break with a Surfing Pikachu"
```
sylvainpolletvillard, line 3024: maybe we can reuse the Diver title since its basically the same ?

app/core/abilities/abilities.ts
```diff
@@ -12478,17 +12492,29 @@ export class BoomBurstStrategy extends AbilityStrategy {
   ) {
     // Deal 60 special damage to all adjacent units including allies
     super.process(pokemon, board, target, crit)
-    const damage = 60
```
sylvainpolletvillard, line 12481: Looks like a MR conflict mistake happened here, not sure what happened

PR #3167, opened 2025-06-06
Fix Unown not deploying in Bot boards

KevinLeoA:
Fix Unowns not deploying in Bot boards that was caused by bots updating before unown deletion check.

Moved bot update to after unown deletion.

Fixes https://discord.com/channels/737230355039387749/1380296663234641971/1380296663234641971


app/rooms/commands/game-commands.ts
```diff
@@ -1726,6 +1725,8 @@ export class OnUpdatePhaseCommand extends Command<GameRoom> {
             }
           })
 
+          // Update Bots after unown deletion so unown in bot boards are not deleted
+          this.state.botManager.updateBots()
```
sylvainpolletvillard, line 1729: to be moved after line 1747

PR #3173, opened 2025-06-12
Update Muscle Band name from Defensive Ribbon

Exeatop:
JavaScript file uploaded as txt to replace the item name in the database: [renameItem.txt](https://github.com/user-attachments/files/20716897/renameItem.txt)


sylvainpolletvillard:
@Exeatop 
I'm not sure about that db update script, what the TempDoc is for

Also we miss the ItemStatistic and GameHustory update, so we'll get two versions of the item in item meta report. This should fix by itself in 2 weeks but it will be broken in the mean time

Exeatop:
> @Exeatop I'm not sure about that db update script, what the TempDoc is for
> 
> Also we miss the ItemStatistic and GameHustory update, so we'll get two versions of the item in item meta report. This should fix by itself in 2 weeks but it will be broken in the mean time

TempDoc is just a throwaway label for the mongoose model; it's not for reuse. 

I can add a script for histories if you'd like, but I don't know how item-statistic is populated, so I can't do that one.

PR #3183, opened 2025-06-19
punching glove changes

sylvainpolletvillard:
Punching Glove now deals additional physical damage equal to ~~8~~ 10% of the target's max HP, applied after all other computations (crit/fairy spcial damage/steel true damage etc.). It is a separate instance of damage, meaning it will be reduced by Fighting resistance, will give 2 stacks to Muscle Band, and won't benefit from other damage boosts like Fairy special damage bonus.

https://discord.com/channels/737230355039387749/1385037988181966868

also added onHitEffect for items refactor


keldaan-ag:
👍

PR #3196, opened 2025-06-26
Fix heal scaling on life drain abilities

KevinLeoA:
Some life drain abilities (Bite, leech life, dream eater) does not have heal % scaling on their description but do so in the code. This changes their effect to match the description.

If this is a description issue instead of effect issue, then ignore this change.


KevinLeoA:
I didn't remove suction healing AP scaling on heal in this PR. If we want to remove that, then can remove it together with changing the ability description.

PR #3200, opened 2025-06-27
Update translations pt

Neyzim:
I've tried to not just translate the text, but also localize the content in order to make it more understandable to native pt speakers

Preview the messages on https://inlang.com/github.com/Neyzim/pokemonAutoChess .


sylvainpolletvillard:
Thank you !

PR #3236, opened 2025-07-08
change darmanitan zen

sylvainpolletvillard:
Change Darmanitan Zen mode: now can attack at range, but with lower attack and speed. Attacks deal special damage, scale with AP and heal for 100% of damage dealt.

Also implemented OnDamageReceivedEffect


app/public/dist/client/locales/fr/translation.json
KtourzaJeremy, line 1: Typo on TRANSE description: "HP amx" -> "HP max"

PR #3297, opened 2025-08-20
add roggenrola line

keldaan-ag:


app/core/abilities/abilities.ts
```diff
@@ -13024,6 +13026,55 @@ export class SpinOutStrategy extends AbilityStrategy {
...
+
+    for (let i = 0; i < numberOfRocks; i++) {
+      const randomEnemy = pickRandomIn(enemies)
+      if (randomEnemy) {
+        const adjacentCells = board.getAdjacentCells(
+          randomEnemy.positionX,
+          randomEnemy.positionY,
+          true
+        )
+        const targetCell = pickRandomIn(adjacentCells)
```
sylvainpolletvillard, line 13054: if i understood correctly, the ability would do zero damage if the target is the only enemy remaining ?

PR #3305, opened 2025-08-25
Update Icy and Heat rocks to include luck.

shadjensen:
Updated the translation files to include luck in the description of Icy and Heat rocks, which is already how they interact in the game but currently isn't being shared with the players.


sylvainpolletvillard:
Thank you !

PR #3319, opened 2025-09-04
ogerpon rework as legendary

sylvainpolletvillard:
Rework Ogerpon ; now legendary. Masks have new effects and higher base stats. Ogerpon loses its mask when its shield is depleted.

Added OnShieldDepletedEffect

https://discord.com/channels/737230355039387749/1404234676591399044


keldaan-ag:
👍

PR #3331, opened 2025-09-13
quickfix - OnSpawnEffect dishes

tegberen:
Dishes like Fruit Juice, Binding Mochi, Tea etc. that activate on spawn are not activating properly at the start of the battle.


sylvainpolletvillard:
Are you sure about that ? OnSpawn effects are supposed to be triggered before consuming the meal. Did you test it on your side ?

PR #3337, opened 2025-09-15
animation-fix for orientedProjectile 

tegberen:
Some abilities like: Spirit Shackle , Dragon Breath, Psycho Cut etc. that use the orientedProjectile function are not rotating correctly the sprites for all its 8 orientations.
Visually good to see with the Spirit Shackle Ability: Ability-Sprite orientation stays  "Down-Left" for all other orientations too

- added orientation check in the if-clause so it only runs for Spiky Shield Ability, otherwise ignored

### DIFF
--- a/app/public/src/game/components/abilities-animations.ts
+++ b/app/public/src/game/components/abilities-animations.ts
@@ -589,7 +589,7 @@ const orientedProjectile: AbilityAnimationMaker<
     oy + dy * (options.distance ?? 8),
     args.flip
   )
-  if (options?.oriented) {
+  if (options?.oriented && options.orientation) {
     options.rotation = angleBetween([dx, -dy], [0, 0]) + (options.rotation ?? 0)
     delete options.oriented
   }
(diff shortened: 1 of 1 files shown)


sylvainpolletvillard:
> added orientation check in the if-clause so it only runs for Spiky Shield Ability, otherwise ignored

i dont understand that change, why should we check for orientation information to apply oriented projectile based on units position ?

tegberen:
At the moment, all abilities that use orientedProjectile, their sprites stay in their default facing direction. 

https://github.com/user-attachments/assets/78f0af20-25a8-441a-b1ee-d6864e9337b6

`if (options?.oriented) {
    (...)
    delete options.oriented }`
When using the flag `oriented: true` it gets deleted in the if clause.

So further in the code when calling `tweenAnimation`, it no longer has `oriented: true`.

https://github.com/user-attachments/assets/7abf538a-ecf5-4223-9608-a3eff0d2316b

Abilities like Spiky_Shield uses the `orientation` parameter to calculate the sprites for each direction. So i just added the `&& options.orientation`. So the if-clause applies correctly for Spiky Shield. For the other abilites, the `tweenAnimation` handles the rotation of the sprites correctly, so they can skip the if-condition.

https://github.com/user-attachments/assets/1cb3dd74-b3c9-4065-8430-4cf0cec74e73

PR #3363, opened 2025-09-26
spite and grudge

tegberen:
potential one less computation for SpiteAbility, since the `addPP` function already checks for the case where more mana is burned than the target has. The crit and ability power is also applied within `addPP`

Dividing the `ppDrain ` number with number of allies. At the moment each unit adjacent will get the full 40pp. the ability states redistributing

Remove  the _flat_ ap scaling from GrudgeAbility

### DIFF
--- a/app/core/abilities/abilities.ts
+++ b/app/core/abilities/abilities.ts
@@ -13604,12 +13604,7 @@ export class SpiteStrategy extends AbilityStrategy {
     crit: boolean
   ) {
     super.process(pokemon, board, target, crit, true)
-    const ppDrain =
-      ([20, 40, 60][pokemon.stars - 1] ?? 60) *
-      (1 + pokemon.ap / 100) *
-      (crit ? pokemon.critPower : 1)
-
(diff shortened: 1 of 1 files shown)


sylvainpolletvillard:
looks goods to me. not sure where this + pokemon.ap on grudge damage comes from

app/core/abilities/abilities.ts
```diff
@@ -13633,7 +13628,7 @@ export class SpiteStrategy extends AbilityStrategy {
             targetX: ally.positionX,
             targetY: ally.positionY
           })
-          ally.addPP(ppDrain, pokemon, 0, crit)
+          ally.addPP(ppDrain/(adjacentAllies.length), pokemon, 1, crit) //divide by number of allies to redistribute
```
keldaan-ag, line 13631: If you change the ability by dividing by the number of allies to redistribute, then you must also update the description.

tegberen: So the division by the number of allies is made to match the current description. In english 'redistribute' means to share something again without increasing the total

PR #3365, opened 2025-09-26
Add an icon for time value (ND) on descriptions

KtourzaJeremy:
Proposal: add an icon for durations similar to the AP icon for the damages.

pac


sylvainpolletvillard:
ND is not linked to time, its a number of decimals

I also dont think such an icon is useful

KtourzaJeremy:
Oh ok no problem, I get it ! 
I was thinking of using this clock icon mostly to replace all the times you used the AP icon for durations.

PR #3379, opened 2025-09-30
Display pokemon names when hovering portraits/avatars

brunopma:
I've been missing this feature specially on the game history section, where I want to hover the pokemons from my team and see their names on the portraits, like the synergies.
image


sylvainpolletvillard:
This code injects a function in a template string, it doesn't work, which makes the assume it has not been tested. Please test your code before making a PR

PR #3419, opened 2025-10-04
add seasonal forms for deerling

sylvainpolletvillard:
Deerling line has now 4 seasonal forms (Spring, Summer, Autumn, Winter) that change appearance and type based on the current season (in Northern Hemisphere)


keldaan-ag:
👍

PR #3427, opened 2025-10-05
Trenches for fully dug ground rows

sylvainpolletvillard:
Ground fully dug holes in a row are nos displayed as trenches instead of individual holes

https://discord.com/channels/737230355039387749/1424349013695533196/1424349013695533196

chrome_w7CriF7uNG

image


keldaan-ag:
👍

PR #3672, opened 2026-02-14
TM rework

sylvainpolletvillard:
- Human HM have been removed, and TM are now split into 3 categories: Bronze, Silver and Gold ; each category matching one synergy level of Human synergy. Removed TM Taunt. New TMs: Charge, Focus Punch, Hyper Beam, Protect, Skill Swap
- TMs are now unlearned and regained dynamically based on Human synergy level
- Nerf Human healing: 25/40/60% → 25/35/50%


app/core/abilities/abilities.ts
```diff
@@ -16007,6 +16007,134 @@ export class WaveSplashStrategy extends AbilityStrategy {
...
+            cell.value.handleSpecialDamage(
+              damage,
+              board,
+              AttackType.SPECIAL,
+              pokemon,
+              crit
+            )
+          }
+        })
+        pokemon.status.triggerFatigue(3000, pokemon)
```
ecustWallace: In doc it says the fatigue time for hyperbeam is 5s, but here it's 3s. Wanna make sure they are aligned.

app/core/abilities/abilities.ts
```diff
@@ -16007,6 +16007,134 @@ export class WaveSplashStrategy extends AbilityStrategy {
...
+  ) {
+    super.process(pokemon, board, target, crit)
+    pokemon.skill = target.skill
+    pokemon.maxPP = target.refToBoardPokemon
+      ? target.refToBoardPokemon.maxPP
+      : target.maxPP
+    if (pokemon.refToBoardPokemon) {
+      pokemon.refToBoardPokemon.skill = target.skill
+    }
+    AbilityStrategies[target.skill].process(pokemon, board, target, crit)
```
ecustWallace: Just think of an edge case: will it cause infinite loop if the opponent is also equipped with skill swap?

PR #3818, opened 2026-04-10
Allow Gourmet pokemon to Dig

shadjensen:
At present, pokemon holding the chef's hat cannot dig. This means cooking pot ground pokemon or explorer kit gourmet pokemon cannot dig. I could not find an explanation for this, and doesn't seem to align with player expectations. 

If this is intended, I apologize but I assumed it was a relic of some previous implementation that should no longer be here.


sylvainpolletvillard:
This is intended. Cooking and digging both cause an animation, this would be weird if they could do both at the same time.

PR #3888, opened 2026-04-22
refactor: replace hardcoded magic numbers in ability descriptions

keldaan-ag:
…h semantic placeholders


sylvainpolletvillard:
I get the intention to ensure descriptions always match the reality in code, but it adds so much clutter and additional work for both devs and translators... is it really worth it ?

PR #3930, opened 2026-05-16
fix fairy wands damage

tegberen:
- special damage double crit on critical hits
- apply fairy damage to non-fairy allies


sylvainpolletvillard:
Thank you for your vigilance, Tegberen <3

PR #3949, opened 2026-06-02
Handle Crit Reduction for Effect Abilities

shadjensen:
Updated abilities that apply effects to their user to be reduced by crit reduction if spell crit is enabled. Currently abilities like Teleport are not reduced by crit reducing effects such as rocky helmet

### DIFF
--- a/app/core/pokemon-state.ts
+++ b/app/core/pokemon-state.ts
@@ -91,6 +91,10 @@ export default abstract class PokemonState {
         }
       }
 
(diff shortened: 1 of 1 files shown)


shadjensen:
This should now be up to date with the critical reductions mentioned here: https://discord.com/channels/737230355039387749/1511246856896647288/1511343117947768922

I made this to merge into prod assuming there might be a few more hotfixes with the new patch release but it can wait for the next patch too if we think the period for hotfixes is over.

PR #3950, opened 2026-06-02
Reduce Spell Crit by Augurite

shadjensen:
Reduce all special damage on crit by black augurite, per the conversation: https://discord.com/channels/737230355039387749/1511246856896647288/1511343117947768922

It is worth noting that this doesn't reduce Teleport damage twice, because this code only triggers if crit is true and `handleSpecialDamage` is called with `crit = false` after leaving the attack method in `pokemon-state.ts`

This bug is reported here: https://discord.com/channels/737230355039387749/1511453438460821704


sylvainpolletvillard:
Thank you !

PR #3974, opened 2026-06-16
Fix Minior Nomicon interaction

shadjensen:
Currently because of the placement of pokenomicon's special defense drop being in handleSpecialDamage, some instances of special damage don't trigger it. One such trigger is Minior, which probably needs to be changed to use the other call but in the meantime this refactor seems preferred to doing a wide review and possibly missing edge cases. 

https://discord.com/channels/737230355039387749/1516308277787623536


sylvainpolletvillard:
I'm going to refactor it as an OnDamageDealt effect to simplify code

sylvainpolletvillard:
closed in favor of https://github.com/keldaanCommunity/pokemonAutoChess/commit/5180ab43eb60ac7394109dedde6f26581e5704cf ; thanks for your help shadjensen

PR #3980, opened 2026-06-21
Fix Baneful Bunker poisoning self instead of attacker

Phoenix-Starlight:
Incorrect unit is referenced, instead of poisoning the attacker as the description says, it poisons itself instead.


sylvainpolletvillard:
Thank you !

PR #4007, opened 2026-06-30
Fix player cycling hotkeys to follow displayed rank order

Cinnlite:
## Summary

This updates the previous/next player board hotkeys so they cycle through players in the same order shown in the right-side player list.

Previously, the hotkeys used the raw player map order, which could differ from the displayed player order. This made Page Up / Page Down feel inconsistent, unintuitive and confusing during games.

## Changes

- Sort players by `rank` before cycling with previous/next player hotkeys
- Dead players are still included in the cycle
- Add a guard so cycling does nothing if the current spectated player cannot be found for some reason

P.S. Not sure under which category to add the change in the patch notes. I added it under bugfix for now.

### DIFF
--- a/app/public/dist/client/changelog/patch-6.11.md
+++ b/app/public/dist/client/changelog/patch-6.11.md
@@ -76,6 +76,8 @@ Added a fourth tier (or fifth tier ?) of ability power per unit star level to al
 
 # Bugfix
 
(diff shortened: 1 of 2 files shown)


sylvainpolletvillard:
The current behavior is intended, see https://discord.com/channels/737230355039387749/1492943052052893746

PR #4020, opened 2026-07-08
In-client replay viewer + recorder

brobinett:
Record your own matches and watch them back in-client at `/replay`. Recording is an always-on (opt-out) background tap of the client's own game stream. Playback re-runs the recorded Colyseus state through the normal game client, so replays render with the real board, combat, and UI. Nothing is re-simulated. Files are small binary `.colreplay` blobs kept on-device (OPFS). The last few games are auto-kept, or you can download one to share.

## Commits
Seven logical commits, each of which builds on its own (CI: `npm run build` + biome lint):
1. **`bindRoom` extraction.** A no-op refactor of the game page's per-room bindings into a function, so it can be re-bound to a fresh room on a seek. Behavior is unchanged; `git show -w` shows just the wrapper, the call, and the room→parameter rename.
2. **`.colreplay` v1 format + readers.** The binary `CLRP` container/encoder/decoder, plus promoting `@colyseus/msgpackr` to a direct dependency.
3. **Recorder.** The WebWorker + OPFS streaming writer and the no-loss flush core.
4. **Replay playback core.** The `ReplayRoom` decoder (`replay-room.ts`), which plays a recorded match's state stream through the normal client. Nothing is re-simulated.
5. **Event-log derivation.** `replay-index.ts`, `replay-combat-scan.ts`, and `replay-event-format.ts`, which diff the decoded state frame-by-frame into the semantic event stream (see "The event log"), plus the `replay.eventlog.*` row/phase strings they localize against. Import-pure (no Phaser/React/Redux) and Node-testable, so it can be reviewed and exercised on its own.
6. **Viewer.** The `/replay` route and UI, the event log, and the viewer's `replay.*` locale keys (controls, library, skew, error boundary).
7. **Dispatch the current shop on bind.** A one-line live fix (a blank shop when you reconnect to your own in-progress game), split out so it can be reviewed and reverted independently of the replay feature.

## Reviewing this PR (suggested reading order)

This is a large diff, but most of it is UI, CSS, and localization. The logic worth scrutinizing is in a few files:

1. **The format.** `replay-format.ts` (the `CLRP` container: what we persist and how it decodes). This is the contract everything else is built on.
2. **The recorder.** `recorder.ts`, `recorder-flush-core.ts`, `opfs-replay-writer.ts`, `recorder.worker.ts`. How we tap the stream, batch frames, and write them to disk without blocking the render thread or losing frames on a slow write.
3. **Playback.** `replay-room.ts` decodes frames back into the client. `replay-index.ts` is the event-log derivation engine (see "The event log" below), with `replay-combat-scan.ts` and `replay-event-format.ts` feeding it; `replay.tsx` is the viewer wiring.
4. **The live-file edits.** Small, but the highest-scrutiny part, because they touch code that runs in normal play. Each is listed with its justification under "Changes to existing game files" below.

Everything else is presentational and low-risk: `replay-ui.css` / `replay-event-log.css`, the `replay.*` locale keys, and the event-log and controls components. The `game.tsx` diff looks large, but most of it is the `bindRoom` extraction with no behavior change (its own commit; verify with `git show -w`).

## How recording works
An always-on tap of the client's own inbound game stream (the same messages the client already receives, so no new server load and no other players' hidden state). Frames are captured on the render thread and handed to a WebWorker that streams a binary `.colreplay` to OPFS. Only about one second of in-flight frames sit in RAM, never the whole match. Cost is roughly 3-4 MB per game on disk; the last few games are auto-kept and older ones pruned. Writes are crash-durable: the file is streamed to disk, not held in memory until the match ends. Recording is single-POV and private: it only ever sees what your client saw, and the file never leaves your device unless you download and share it. It is on by default and opt-out, via a checkbox in the Interface options.

## File format
`.colreplay` is a binary `CLRP` container. A small header (magic, version, and a length-prefixed JSON metadata block: game version/commit, room, POV uid, timestamp) is followed by a stream of length-prefixed frame records read to EOF. Three frame kinds store the raw inbound Colyseus bytes (handshake, state, patch) with no base64 or JSON wrapper. A fourth stores `ROOM_DATA` messages, whose payloads arrive already-decoded and are re-encoded with `@colyseus/msgpackr` (the same msgpack codec Colyseus uses to unpack them). Integers are little-endian, and per-frame timestamps are stored as varint deltas. Each frame is independently decodable, since every msgpack payload is self-contained. Seeking depends on it.

An optional match-summary footer at EOF (`CLTR` magic, a `u32` length, and a small JSON blob: final placement, deployed team, POV name) lets the replay library and the file picker render a preview row (rank plus team portraits) from a cheap tail-read, without decoding any frames. It is a footer rather than a header field because the final team is only known when the match ends, long after the streaming writer emits the header at game-join. A recording without it (any older file) just shows no preview, and the decoder bounds the frame loop to exclude it.

## The event log
The event log is derived, not recorded. The `.colreplay` stores only raw Colyseus state, the same bytes the client received; nothing in it says "player X bought Y" or "rerolled." `buildReplayIndex` (`replay-index.ts`) re-decodes the whole transcript once and reconstructs a semantic event stream by diffing consecutive state frames: a shop slot emptying alongside a board change is a buy, a shop refreshing with no board change is a reroll, a board unit leaving with no add is a sell, an added unit that is the evolution of a removed one is an evolve, and so on. The same pass derives round results, level-ups, synergy-tier activations, region changes, item moves (equip / craft / unequip), and per-unit combat status and stat changes (`replay-combat-scan.ts`); `replay-event-format.ts` then localizes each row at render.

This diff-based approach is why `replay-index.ts` is the largest logic file and the one most worth scrutinizing: it is a stack of heuristics, one per event type, each encoding a small "why this state delta means this event" judgment. A few properties follow from it:
- Most events derive for every player, since opponent boards, economy, and synergies all sync, so the log can be filtered per player. Shop-only signals (buy vs. reroll vs. a free "gained") need the shop, which `@view` only syncs to the POV, so those are POV-only.
- Combat casts and damage are camera-scoped: `broadcastToSpectators` only reaches the board the recorder was watching. Status and stat changes are recovered for every board, which is what the filter chips split on.
- It is best-effort. A heuristic can mislabel an unusual sequence, so the log is an optional, filterable panel rather than a source of truth; the board and combat playback is authoritative.

The engine is import-pure (no Phaser, React, or Redux) and Node-testable, so it can be reviewed and exercised on its own.

## Versioning
A replay records Colyseus state and messages and replays them through the normal client; combat is never re-simulated. So balance patches don't affect replays. Only structural changes (schema, message protocol, assets) can. Decoding is self-describing, since the schema rides in the recording's handshake, so what drifts across such a change is the rendering, and only for an old saved `.colreplay` opened in a newer build (the auto-kept recent recordings are always current). Each file is stamped with its build (game version plus `assetsVersion`); on a mismatch the viewer warns instead of failing silently. Replays are reliable within a version and best-effort across versions.

## Changes to existing game files
The viewer reuses the unchanged game page (the `/replay` route mounts the real `<Game/>` in a read-only mode), so most of the code is additive. The edits that touch code running in normal play are few; apart from the recorder's own always-on wiring (`index.tsx` install and the `lobby.tsx` drain below), each is either inert live or a deliberate latent-bug fix:

- **`game.tsx`.** (a) `bindRoom` extracted to a named function (pure refactor, its own commit; behavior unchanged). (b) The player's shop is dispatched up-front on bind, which also fixes a live blank-shop-on-reconnect. (c) A `GAME_END` guard, (d) `spectate = true`, and (e) the camera-follow guard are all gated on `roomId === "replay"`, so they are no-ops in live play.
- **`GameStore.addPlayer`.** An idempotent `findIndex` dedup (no-op live), plus a `synergies` re-upgrade that runs on every live add and prevents a latent `.entries()` crash when hovering a player before `setSynergies` has run.
- **`game-pokemon-portrait.tsx`.** A per-index base64 portrait cache (perf), and a `scene.textures.exists()` gate that also fixes a live bug: `getBase64` returns `""` for an absent texture, which the old `?? ` turned into a broken `url("")`.
- **`index.tsx`.** Installs the recorder at module load, gated on the opt-out preference. The recorder reads the client's own inbound stream by wrapping the Colyseus SDK's decode seam (`SchemaSerializer` + `Room.dispatchMessage`) at startup. That's the only way to capture the schema-bearing join handshake, which decodes before the room is exposed to the app and isn't on any public hook. It only reads, is best-effort (try/catch, falls through to the original, disables with a log if a future SDK bump moves the seam), and skips non-game rooms so watching a replay doesn't re-record it. This is the one place the feature reaches into vendored internals.
- **`lobby.tsx`.** On return to the lobby, flushes the finished game's in-memory tail to disk and releases the recorder's room reference (`resetActiveGameRoom`). Runs every live session; the drain is bounded so it can't stall the lobby transition.
- **`game-scene.ts` / `wanderers-manager.ts` / `loading-manager.ts`.** Small faithful refactors: tracked timers cleared on teardown, and a `textures.exists()` guard on (re)load. The `loading-manager` guard is unconditional (it skips re-loading an already-cached atlas in any game); it is behavior-preserving because textures are destroyed between matches.

## New dependency
**`@colyseus/msgpackr`** is promoted from a transitive dependency to a direct one. It is the msgpack codec the Colyseus SDK already uses to (de)serialize `ROOM_DATA`. The recorder re-encodes decoded `ROOM_DATA` payloads with the same codec, so playback feeds byte-identical messages back to the client.

## Limitations of client-side capture
- **Single POV.** A recording only ever contains what your client saw. You see your own shop and board, and opponents only as much as the game already synced to you.
- **Disconnects leave a gap.** If your client drops mid-match, the frames it didn't receive aren't in the recording; playback resumes from the reconnect.
- **Playback is best-effort on timing, not a re-simulation.** Speed controls are approximate, animations render at their native rate, and audio is regenerated live by the client.


app/public/src/game/components/loading-manager.ts
```diff
@@ -38,6 +38,11 @@ export default class LoadingManager {
   async preload() {
     const scene = this.scene
     scene.load.xhr.timeout = 5000 // help avoiding failed loading of assets when server is overloaded
+
+    // skip cached multiatlases; a replay seek re-preloads the same keys
```
sylvainpolletvillard: Phaser’s Scene Loader is not supposed to reload the actual network files if they are already present in the global Cache. I'd like to know more about the issue you met here

app/public/src/game/components/wanderers-manager.ts
```diff
@@ -31,13 +32,31 @@ List of wanderers:
 export default class WanderersManager {
   scene: GameScene
 
+  // deferred wanderer timers; Phaser doesn't cancel raw setTimeouts on shutdown, so dispose() clears them on a replay seek
+  private timers = new Set<ReturnType<typeof setTimeout>>()
```
sylvainpolletvillard: this is likely to be useful at many other places of the codebase, we should make this a reusable helper class

sylvainpolletvillard: there's also the Phaser Clock API that can do delayed calls and auto dispose on scene destroy
https://docs.phaser.io/api-documentation/class/time-clock

app/public/src/pages/component/main-sidebar/main-sidebar.tsx
```diff
@@ -131,6 +133,12 @@ export function MainSidebar(props: MainSidebarProps) {
     )
   )
   function onClickLeave() {
+    // a replay exits straight to the lobby (the live leave() would hit the /after flow)
+    if (isReplayRoom(rooms.game)) {
+      ;(rooms.game as unknown as { pause?: () => void }).pause?.()
```
sylvainpolletvillard: this type assertion can be removed by adding a type predicate to the isReplayRoom function

app/public/src/pages/component/options/game-options-modal.tsx
```diff
@@ -212,6 +212,14 @@ export default function GameOptionsModal(props: {
               label={t("options.colorblind_mode")}
             />
           </p>
+          <p>
```
sylvainpolletvillard: Can you move this to its own tab in game options ? I dont think it has anything to do with Interface, and i'd like to add more options related to replays.

app/public/src/pages/component/replay/replay-controls.tsx
```diff
@@ -0,0 +1,402 @@
...
+  try {
+    const s = localStorage.getItem(POS_KEY)
+    return s ? JSON.parse(s) : null
+  } catch {
+    return null
+  }
+}
+
+// inline SVG control icons (24×24, currentColor to inherit text color + disabled dim); matches the game's own SVG-asset buttons
+const RcIcon = ({ d, d2 }: { d: string; d2?: string }) => (
```
sylvainpolletvillard, line 42: I think we will likely move to external SVG files for more complex icons and customability per theme. You can keep it like this for the context of this PR

app/public/src/pages/component/replay/replay-controls.tsx
```diff
@@ -0,0 +1,402 @@
...
+  nextStage,
+  prevPhase,
+  prevStage,
+  segmentAt,
+  segmentBandKind,
+  type ReplayIndex
+} from "../../../game/replay-index"
+import { phaseWord } from "../../../game/replay-event-format"
+import type { ReplayRoom } from "../../../game/replay-room"
+import "./replay-ui.css"
```
sylvainpolletvillard: this CSS import is repeated on many components, it should either be imported once on common parent or split in several indidual files per component

app/public/src/pages/game.tsx
```diff
@@ -241,11 +254,17 @@ export default function Game() {
...
+      navigate("/lobby")
+      return
+    }
     const afterPlayers = new Array<IAfterGamePlayer>()
 
     const token = await firebase.auth().currentUser?.getIdToken()
 
     if (gameContainer && gameContainer.game) {
+      // textures die with the TextureManager here; drop the portrait cache or the next game shows stale sprites
+      clearPortraitBase64Cache()
```
sylvainpolletvillard: this should be called on an event listener on scene destroy event instead, to not have to manually call it everywhere we quit the scene.

app/public/src/stores/GameStore.ts
```diff
@@ -113,7 +113,18 @@ export const gameSlice: Slice<GameStateStore> = createSlice({
       state.specialGameRule = action.payload
     },
     addPlayer: (state, action: PayloadAction<IPlayer>) => {
-      state.players.push(JSON.parse(JSON.stringify(action.payload)))
+      // idempotent by id: a replay seek re-fires onAdd, so replace rather than push (avoids duplicate players)
+      const clone = JSON.parse(JSON.stringify(action.payload)) as IPlayer
```
sylvainpolletvillard, line 117: I don't remember why we do the json clone here, but structuredClone() could also be used to not have to recreate schemas here.

app/public/src/index.tsx
```diff
@@ -12,12 +12,17 @@ import Game from "./pages/game"
...
+import Replay from "./pages/replay"
 import { SpriteDebug } from "./pages/sprite-viewer"
 import TranslationsPage from "./pages/translations"
+import { installRecorder } from "./game/recorder"
 import store from "./stores/index"
 import "./style/index.css"
 import "./theme"
 
+// tap the inbound colyseus stream before any game room joins, so a match can be recorded and downloaded
+installRecorder()
```
sylvainpolletvillard, line 24: the recorder should not be installed at all if recordReplays preference is set to false. This is safer in case there's a bug in Colyseus patching code, as it allows players to fix it just by disabling the option.

sylvainpolletvillard (changes requested):
Thank you ! This code will be hard to maintain for us, especially the recorder code. I hope it won't prevent us to make evolutions in the game while not breaking this replay system.

I have a few remarks after review, please have a look

sylvainpolletvillard:
image
There's also some blatant CSS issues like this to be fixed, please try in various screen resolutions

sylvainpolletvillard:
Continuing the testing, I don't manage to watch a replay due to Firebase not being initialized in replay view apparently:
image

brobinett:
> Thank you ! This code will be hard to maintain for us, especially the recorder code. I hope it won't prevent us to make evolutions in the game while not breaking this replay system.

On maintainability, I'm hopeful that this won't have much of an upkeep burden. The recorder just records whatever gets sent over the wire, so pretty much all balance patch stuff replays for free. Structural schema changes also don't need code changes; recordings carry their own reflection schema and build stamp, old files show a warning. An SDK bump that moves the decode seam could idle the recorder, but it would announce itself. There are basically two things that actually will need updates. The first is the event log which is derived per mechanic, so a new mechanic won't have a labeled row until someone adds one. Replays will still play back, so it just degrades, it doesn't break. The other is that the viewer replays through the game.tsx bindings, so a refactor there could break playback silently. Good spot for a small playback test if you want one.

brobinett:
> Continuing the testing, I don't manage to watch a replay due to Firebase not being initialized in replay view apparently: image

On the firebase error, that one's on me. I didn't want to set up a firebase account so I've been doing all my development on a patched version that doesn't need it. Two changes, pass uid through scene start data which removes the renderer's firebase dependency, and /replay inits firebase on mount. Passing the uid also fixes a race, even logged in, firebase.auth().currentUser is null synchronously on a hard refresh, so reading it in GameScene.init could hang the viewer, live games currently avoid this because authenticateUser awaits onAuthStateChanged before connecting.

brobinett:
> image There's also some blatant CSS issues like this to be fixed, please try in various screen resolutions

Swept from 390x844 to 3440x1440. Still gets kind of bad looking at phone widths, but I think everything is at least correct now (no overflows, etc). There's a couple pixels of horizontal scroll at phone width that are preexisting.

app/public/src/game/components/loading-manager.ts
brobinett: Seeking in the replay viewer restarts the scene, so without the guard, preload re-queues the six multiatlases. The loader's duplicate-key check doesn't catch multiatlas files, so they re-download and re-decode, then the TextureManager rejects each re-add, six console errors per seek. Playback still works, so the guard is just about that fetch + decode + console errors. In live games, textures are destroyed between matches so this is a no-op.

app/public/src/game/components/wanderers-manager.ts
brobinett: I went ahead with the Phaser clock, but this is a behavior change, so if you don't want that I can go with the other approach. The scene clock pauses while the tab is hidden, whereas setTimeout kept running, so things now resume when the tab comes back instead of firing unseen.

app/public/src/pages/component/main-sidebar/main-sidebar.tsx
brobinett: Done

app/public/src/pages/component/options/game-options-modal.tsx
brobinett: Done. Didn't move the keep-N retention setting off the /replay page, but can if you want.

app/public/src/pages/component/replay/replay-controls.tsx
brobinett: Done

app/public/src/pages/game.tsx
brobinett: Done. I used game's destroy rather than scene shutdown because the textures are game-global and a replay seek restarts the scene while they stay alive. One manual clear is still there, on game entry a ROOM_DELETED/USER_BANNED exit navigates away without destroying the game, so no destroy event fires on that path.

app/public/src/stores/GameStore.ts
brobinett: Tried this, structuredClone doesn't work here. The JSON round-trip goes through Schema.toJSON which collapses MapSchema into the plain object the reducers read. structuredClone skips toJSON and clones the MapSchema internals instead, so player.synergies stops being readable downstream.

app/public/src/index.tsx
brobinett: Done

app/public/src/game/components/loading-manager.ts
sylvainpolletvillard: > The loader's duplicate-key check doesn't catch multiatlas files

I think it's more complex than that. Phaser does do a duplicate-key check, but only for the JSON file asset itself. Then inside the JSON is referred an array of multiple image files, that are injected to the loader queue on the fly. We just happen to have multiatlas with only one texture and the same key than the atlas itself, but we shouldn't take this case for the rule of thumb.

app/public/src/game/components/wanderers-manager.ts
sylvainpolletvillard: I think it's okay, we can still manually call clearPendingEvents in between stage phases so that they don't stack up if user switched tab for too long.

app/public/src/stores/GameStore.ts
sylvainpolletvillard: yeah we probably dont do things right here when mixing Redux state and colyseus schemas. OK to keep this workaround for the context of this PR ; it's too big of a task to fix this here.

sylvainpolletvillard:
Appreciate the quick bugfixes and detailed answers. I'll do another round of tests this weekend.

sylvainpolletvillard:
I still have some issues watching replays linked to maps preloading apparently:
image
image
image

Inside this.load.image() and this.make.tilemap() the scene property is set to `null`. I'm investigating why

sylvainpolletvillard:
The bug is gone and the replay loads much faster if I comment this line: https://github.com/brobinett/pokemonAutoChess/blob/f0821cb70704548100ad071fff87913f40a48d45/app/public/src/pages/replay.tsx#L512

image


I think that unmount code is triggered at the wrong time.

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"egg_named": "Laid {{pkm}} egg",
+				"egg_golden": "Laid golden {{pkm}} egg",
+				"fish": "Fished {{pkm}}",
+				"gained": "Gained {{pkm}}",
+				"got_item": "Got {{item}}",
+				"equip": "Equipped {{item}} on {{unit}}",
+				"unequip": "Unequipped {{item}} from {{unit}}",
+				"craft": "{{c0}} + {{c1}} → {{result}}",
+				"craft_on": "{{c0}} + {{c1}} → {{result}} on {{unit}}",
+				"move": "{{pkm}} → ({{x}},{{y}})",
```
sylvainpolletvillard: Several of those labels in the event log are just symbols and don't seem worth translating. I suggest replacing all symbols to proper english words to justify having a label


```suggestion
				"move": "{{pkm}} moved to ({{x}},{{y}})",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"combatmove": "{{unit}} ({{fromX}},{{fromY}}) → ({{x}},{{y}})",
+				"level": "→ {{level}}",
```
sylvainpolletvillard:
```suggestion
				"level": "Reached level {{level}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"egg_golden": "Laid golden {{pkm}} egg",
+				"fish": "Fished {{pkm}}",
+				"gained": "Gained {{pkm}}",
+				"got_item": "Got {{item}}",
+				"equip": "Equipped {{item}} on {{unit}}",
+				"unequip": "Unequipped {{item}} from {{unit}}",
+				"craft": "{{c0}} + {{c1}} → {{result}}",
+				"craft_on": "{{c0}} + {{c1}} → {{result}} on {{unit}}",
+				"move": "{{pkm}} → ({{x}},{{y}})",
+				"combatmove": "{{unit}} ({{fromX}},{{fromY}}) → ({{x}},{{y}})",
```
sylvainpolletvillard:
```suggestion
				"combatmove": "{{unit}} moved from ({{fromX}},{{fromY}}) to ({{x}},{{y}})",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"level": "→ {{level}}",
+				"synergy": "{{synergy}} {{count}}",
+				"berry_ripe": "{{berry}} ripe",
+				"berry_trees": "Berry trees: {{list}}",
+				"flower": "Flower → {{flower}}",
```
sylvainpolletvillard:
```suggestion
				"flower": "A flower pot evolved: {{flower}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"equip": "Equipped {{item}} on {{unit}}",
+				"unequip": "Unequipped {{item}} from {{unit}}",
+				"craft": "{{c0}} + {{c1}} → {{result}}",
+				"craft_on": "{{c0}} + {{c1}} → {{result}} on {{unit}}",
+				"move": "{{pkm}} → ({{x}},{{y}})",
+				"combatmove": "{{unit}} ({{fromX}},{{fromY}}) → ({{x}},{{y}})",
+				"level": "→ {{level}}",
+				"synergy": "{{synergy}} {{count}}",
+				"berry_ripe": "{{berry}} ripe",
+				"berry_trees": "Berry trees: {{list}}",
```
sylvainpolletvillard: Seems to be the incorrect label,  I see "berries" in code
```suggestion
				"berries": "New berry trees: {{list}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"flower": "Flower → {{flower}}",
+				"wanderer": "Wandering {{pkm}}",
```
sylvainpolletvillard:
```suggestion
				"wanderer": "A wandering {{pkm}} enters the scene",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"pick": "Picked {{chosen}}",
+				"pick_over": "Picked {{chosen}} over {{alternatives}}",
+				"town": "Town: {{npc}}",
```
sylvainpolletvillard:
```suggestion
				"town": "Town encounter: {{npc}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"rule": "Special rule: {{rule}}",
+				"weather": "Weather: {{weather}}",
+				"status": "{{unit}} · {{status}}",
```
sylvainpolletvillard:
```suggestion
				"status": "{{unit}} got {{status}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"synergy": "{{synergy}} {{count}}",
+				"berry_ripe": "{{berry}} ripe",
+				"berry_trees": "Berry trees: {{list}}",
+				"flower": "Flower → {{flower}}",
+				"wanderer": "Wandering {{pkm}}",
+				"pick": "Picked {{chosen}}",
+				"pick_over": "Picked {{chosen}} over {{alternatives}}",
+				"town": "Town: {{npc}}",
+				"rule": "Special rule: {{rule}}",
+				"weather": "Weather: {{weather}}",
```
sylvainpolletvillard:
```suggestion
				"weather": "Weather changed to {{weather}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"status": "{{unit}} · {{status}}",
+				"status_poison": "{{unit}} · Poisoned ({{count}})",
```
sylvainpolletvillard:
```suggestion
				"status_poison": "{{unit}} got poisoned ({{count}} stacks)",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"stat": "{{unit}} {{stat}} {{delta}}",
```
sylvainpolletvillard: make it two labels for buffs and debuffs

for the buff

```suggestion
				"stat": "{{unit}} gained {{delta}} {{stat}}",
```

for the debuff
```suggestion
				"stat": "{{unit}} lost {{delta}} {{stat}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"cast": "{{caster}} · {{skill}}",
```
sylvainpolletvillard:
```suggestion
				"cast": "{{caster}} casted {{skill}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"cast_at": "{{skill}} @({{x}},{{y}})",
```
sylvainpolletvillard:
```suggestion
				"cast_at": "Casted {{skill}} at ({{x}},{{y}})",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"dmg": "{{src}} {{amount}} {{type}} → {{target}}",
```
sylvainpolletvillard:
```suggestion
				"dmg": "{{src}} dealed {{amount}} {{type}} to {{target}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"heal": "{{src}} +{{amount}} → {{target}}",
```
sylvainpolletvillard:
```suggestion
				"heal": "{{src}} healed {{amount}} HP to {{target}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"heal_shield": "{{src}} +{{amount}} shield → {{target}}",
```
sylvainpolletvillard:
```suggestion
				"heal_shield": "{{src}} gave {{amount}} shield to {{target}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"board_effect": "{{effect}} at ({{x}},{{y}})",
+				"life_lost_one": "{{count}} life lost",
+				"life_lost_other": "{{count}} life lost",
+				"income": "+{{total}}g ({{parts}})",
```
sylvainpolletvillard:
```suggestion
				"income": "+{{total}} gold ({{parts}})",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"income_plain": "+{{total}}g",
```
sylvainpolletvillard:
```suggestion
				"income_plain": "+{{total}} gold",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"income_base": "{{n}} base",
+				"income_interest": "{{n}} interest",
+				"income_streak": "{{n}} streak",
+				"placed": "Placed #{{rank}}",
```
sylvainpolletvillard:
```suggestion
				"placed": "Placed {{rank}}",
```

use getRankLabel function to show the final rank

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"region_maps_one": "{{count}} region map",
+				"region_maps_other": "{{count}} region maps",
+				"game_start": "Game start",
+				"game_over": "Game over",
+				"cooked": "Cooked {{dishes}}",
+				"dig": "Dug{{where}}{{found}}",
+				"dig_site": " ({{x}},{{y}}) depth {{depth}}",
+				"dig_treasure": " · {{item}}",
```
sylvainpolletvillard:
```suggestion
				"dig_treasure": ", found {{item}}",
```

app/public/dist/client/locales/en/translation.json
```diff
@@ -3878,6 +3878,170 @@
...
+				"board_effect": "{{effect}} at ({{x}},{{y}})",
+				"life_lost_one": "{{count}} life lost",
+				"life_lost_other": "{{count}} life lost",
+				"income": "+{{total}}g ({{parts}})",
+				"income_plain": "+{{total}}g",
+				"income_base": "{{n}} base",
+				"income_interest": "{{n}} interest",
+				"income_streak": "{{n}} streak",
+				"placed": "Placed #{{rank}}",
+				"region_maps_one": "{{count}} region map",
```
sylvainpolletvillard: This region_maps event doesn't seem really useful

app/public/src/game/components/loading-manager.ts
sylvainpolletvillard: Since we don't have a better workaround for now, I'm OK to keep it but please rename it to loadMultiAtlas and add code comments explaining what we just talked about;

app/public/src/pages/component/replay/recorder-endgame.tsx
```diff
@@ -0,0 +1,56 @@
...
+      clearInterval(id)
+    }
+  }, [roomId])
+
+  if (!room || !info || info.frames === 0) return null
+
+  return (
+    <div className="recorder-endgame">
+      <button
+        className="bubbly blue"
```
sylvainpolletvillard:
```suggestion
        className="bubbly dark"
```

app/public/src/pages/component/replay/replay-event-log.css
```diff
@@ -0,0 +1,357 @@
+/* replay event-log panel; matches the game's .my-container shell. default position is under the DPS meter
+   (same right offset, clearing the right-edge portraits); draggable by its header and resizable, both persisted */
+
+.replay-eventlog {
+  /* per-category accent colors: one source for both the filter chip fill and the row type text */
+  --replay-eventlog-cat-combat: #ff9a8f;
```
sylvainpolletvillard: these color variables should be moved to colors.css and should use existing color variables when appropriate

app/public/src/pages/game.tsx
```diff
@@ -430,6 +450,8 @@ export default function Game() {
       initialized.current = true
 
       gameContainer = new GameContainer(container.current, uid, room)
+      // replay is a spectate session: startGame keys "self" off the signed-in user, not the recorded pov
+      if (isReplayRoom(room)) gameContainer.spectate = true
```
sylvainpolletvillard: change the spectate property in the GameContainer constructor instead

app/public/src/pages/replay.tsx
```diff
@@ -0,0 +1,912 @@
...
+  document.body.classList.add("replay-mode")
+  const block = (e: Event) => {
+    const target = e.target
+    if (target instanceof Element && target.closest(READONLY_CONTROLS)) {
+      e.stopPropagation()
+      e.preventDefault()
+    }
+  }
+  // suppress native context menu: replay overlays sit outside #game-wrapper (the only node the game blocks), so a capture-phase handler stops right-click escaping to the browser menu
+  const blockContextMenu = (e: Event) => e.preventDefault()
```
sylvainpolletvillard: blocking the entire document is a bad idea. Replace this by a `onContextMenu={(e) => e.preventDefault()}` on the replay.tsx root element, just like what we do for the game wrapper. Also you shouldn't block base click events

app/public/dist/client/locales/en/translation.json
```diff
@@ -4028,6 +4193,7 @@
 		"disable_camera_shake": "No camera shake or flashes",
 		"antialiasing": "Enable anti-aliasing",
 		"colorblind_mode": "Colorblind mode",
+		"record_replays": "Record replays",
```
sylvainpolletvillard:
```suggestion
		"record_replays": "Record game replays automatically in the background",
```

sylvainpolletvillard (changes requested):
After my last fix (please review), everything looks good on the technical side

Second review mainly focusing on labels and UX

app/public/dist/client/locales/en/translation.json
brobinett: Applied all suggestions (with a couple grammatical tweaks), and also swept the rest of the labels with the same treatment.

app/public/src/pages/component/replay/replay-event-log.css
brobinett: Moved as --color-replay-cat-*. Economy is now reusing --color-fg-gold, but the rest didn't have close matches. Feel free to change the colors around, I was just trying to keep things visually distinct.

app/public/dist/client/locales/en/translation.json
brobinett: "Placed 4th place" reads a bit weird so I went with "Finished in 4th place"

app/public/dist/client/locales/en/translation.json
brobinett: I removed it

app/public/src/pages/game.tsx
brobinett: Done

app/public/src/pages/replay.tsx
brobinett: Done

brobinett:
Reviewed the unmount fix. I agree it's a better shape than the wrapper. Only one change, which I went ahead and did. On the replay error catch, leave() isn't getting called so nothing was destroying the Phaser instance anymore. I put the destroy in the boundary's componentDidCatch, so that path should be covered without the need for the wrapper.

app/public/src/game/components/loading-manager.ts
brobinett: Done, with comment.

sylvainpolletvillard (approved):
Good to merge for me ! Thanks for playing along with the review process and congratulations for pushing such a complex feature.

brobinett:
Thanks for all the feedback and collaboration throughout the design and review! I'm excited for this to get into the live game and for people to start using it. If there ends up being more work needed on the feature in the future feel free to send some of it my way.

PR #4032, opened 2026-07-18
[DRAFT] Centralize balance values shared by gameplay and translations

Cinnlite:
## What this PR is trying to solve

Many balance values currently exist once in the gameplay code and then again in every translation.

If someone changes an ability’s damage or duration, they also have to find the same number in every language and update it there. This is tedious and can result in descriptions showing values that no longer match the game.

The goal is to move balance-relevant values into config files and let gameplay and descriptions read from the same source. Translators still control the wording, but numbers are inserted using placeholders such as `{{damage}}` or `{{duration}}`.

After a category has been migrated, changing a balance value should only require changing the config once.

## General approach

The balance configs are separated by the existing categories:

- Abilities
- Items
- Passives
- Statuses
- Synergies
- Effects
- Weathers

These are plain objects in `config/game`.

A regular value can simply be stored as a number:

```ts
{
  freezeDuration: 1.5
}
```

Values that depend on stars, AP or Luck can include that information explicitly:

```ts
{
  damage: {
    valuePerTier: [10, 20, 40, 80],
    apScaling: 1
  }
}
```

Gameplay either reads the number directly or uses the shared balance calculation for tiered or scaling values.

The corresponding translation looks like this:

```text
Deal {{damage}} SPECIAL damage and FREEZE for {{freezeDuration}} seconds.
```

The client displays the configured value using the existing tier, AP, Luck and decimal formatting. The translation editor also checks that every field from the config appears in the translated description.

## Scope

I only want to extract values that are relevant to gameplay balance, such as:

- Damage and healing
- Durations
- Chances
- Stat changes
- Thresholds and caps
- Cooldowns and intervals
- Tiered values
- AP and Luck scaling

Other general UI numbers do not need to be moved.

Patch notes are a separate case. Their numbers describe older versions of the game, so they should not use the current balance configs. The old display handling can remain isolated there.

## Migration order

I want to work through this by category rather than changing everything at once:

1. Shared config, calculation and display logic.
2. Synergies and their tier effects.
3. Board effects, statuses and weathers.
4. Items.
5. Passives.
6. Abilities in smaller batches.
7. Final audit, fixing and cleanup.

The shared implementation is already working. Statuses and weathers have been migrated, and I am currently finishing synergies and all their effects. Several abilities plus one item and passive were migrated earlier to test the approach.

Abilities will be done last because they are by far the largest category and contain more individual calculations and edge cases.

## How each value will be migrated

For every category or ability batch, I plan to:

1. Find the gameplay implementation and every place where the value is used.
2. Record the existing value and formula before changing it.
3. Check whether it includes:
   - Star indexing or tier fallback.
   - AP or Luck scaling.
   - Percentages or flat values.
   - Seconds converted to milliseconds.
   - Rounding.
   - Caps or thresholds.
   - Inherited values from an earlier synergy tier.
   - Multiple applications of the same value.
4. Add the equivalent field to the appropriate config.
5. Replace the gameplay literal with the config value without otherwise rewriting the mechanic.
6. Replace the numbers in descriptions with named placeholders.
7. Check the changes in every available language.
8. Run the migration and build checks.
9. Test representative cases in the game.

If something is ambiguous, I will stop and ask instead of guessing the intended behavior.

## Testing gameplay equivalence

Before migrating a mechanic, I will record its current outputs for a set of representative cases. The same cases will then be checked after the migration.

For a tiered or scaling value, this can include:

- Every supported star or synergy tier.
- A tier above the available array length to verify the fallback.
- AP values such as 0, 50 and 100.
- Luck values such as 0, 50 and 100.
- Relevant boundary conditions, such as exactly 30% HP.
- The maximum value and one value beyond a cap.

The tests need to check the value actually used by gameplay, not only confirm that the config contains the expected number. Otherwise, a test could pass while gameplay was still using an old hardcoded value.

Different mechanics will need different checks:

- Damage and healing: compare the final amount before and after.
- Chances: compare the calculated probability before the random roll.
- Durations: check both the configured seconds and the milliseconds passed to the timer or status.
- Thresholds: test immediately below, exactly at and immediately above the threshold.
- Caps: test the maximum and one value beyond it.
- Tier inheritance: verify that higher synergy tiers for example still include bonuses from earlier tiers when their own config only defines the newly added effect.
- Repeated effects: check both the value and how often it is applied.
- Scaling: check the base result and results with AP or Luck.
- Rounding: use values where `round`, `ceil` or decimal formatting changes the result.

Where possible, I will add a characterization test before changing the implementation and keep the expected result unchanged afterward.

Some effects are tied deeply into the simulation and are harder to isolate. For those, I will compare the old formula with the new config-based calculation and manually test a representative case in-game.

I also want to temporarily change at least one migrated value per category to an obvious test value. This confirms that the same config affects both the tooltip and gameplay. The temporary value will then be reverted.

Tests that remain useful after the migration, such as the shared tier/AP/Luck calculation tests, can stay in the repository. One-time migration comparisons and reports do not need to remain once the work is complete.

## Translation validation

For every locale, I will check that:

- The JSON is valid.
- Every field in the config appears in the description.
- No old `[x,x,SP]`, `[LK]` or `[ND=...]` syntax remains in a migrated category.
- Placeholders still have the correct units, such as `%` or seconds.
- Repeated or reordered numbers were mapped to the correct meaning.
- The rendered English description contains the same values as gameplay.
- Missing localized descriptions correctly fall back to English.

Automated checks can detect missing placeholders, but they cannot always tell whether a placeholder was inserted into the correct part of a translated sentence. I will therefore also try to review the before/after lines and test samples from every language in the UI.

## Handling active translation changes

Since translation contributors are currently active, I do not think it makes sense to repeatedly resolve translation conflicts throughout the migration.

My current idea is:

1. Keep migrated translations during development so the UI can be tested.
2. Finish the gameplay and config changes.
3. Update the branch with the latest master shortly before final review.
4. Take the newest translation wording from master.
5. Reapply the placeholder migration to those fresh translation files.
6. Run the complete translation validation again.
7. Manually inspect ambiguous replacements and known mismatches.
8. Merge relatively soon afterward so another large set of conflicts does not accumulate.

AI can help with the repetitive replacements, but those replacements should follow a known per-description mapping instead of blindly replacing every occurrence of a number.

For example:

```text
WATER_VEIL:
70 → {{statusReductionPercent}}
30 → {{waveHealPercent}}
15 → {{waveDamagePercent}}
14 → {{secondWaveTime}}
```

This matters because the same number can have different meanings, and translated sentences do not always use the same order as English.

## Existing mismatches

The migration has already exposed descriptions that did not match gameplay.

I think the general rule should be:

- If gameplay appears correct but a translation is stale, update the description to match gameplay.
- If gameplay itself appears questionable, preserve it for the migration and report it separately.
- If it is unclear which side is correct, ask before changing either one.
- Intentional balance changes should be documented separately from the mechanical migration.

## Checks before a batch is finished

For every completed batch, I will run:

- TypeScript compilation.
- Client build.
- Existing relevant tests.
- Characterization or comparison tests for the migrated behavior.
- Config-placeholder validation for every locale.
- A search for remaining legacy syntax and hardcoded balance values.
- Manual diff review.
- Manual in-game testing of the riskiest mechanics.

A category is finished when gameplay uses the config, descriptions use the same values, all checks pass and there are no unexplained behavior changes.

There is no honest way to guarantee that a migration of this size can never contain a mistake. The aim is to make every moved value traceable and its before/after behavior equivalent.

AI can help locate usages and perform repetitive edits, but the verification will come from the before/after mappings, automated checks, tests, builds, some manual diff review and in-game testing.

## Expected result

Once a category has been migrated, changing one of its balance values should only require editing its config.

Gameplay and every localized description will then use the new value automatically, without requiring the same numeric edit in every translation file.

## Questions

Before doing the largest remaining categories, I would like confirmation on:

1. Should the migration remain in one large PR with separate commits, or be split into category PRs?
2. Should patch notes permanently keep their isolated legacy number formatting?
3. Is the proposed scope too much and you think we should keep the scope smaller?


app/public/src/pages/component/ability/ability-tooltip.tsx
```diff
@@ -14,7 +15,7 @@ export function AbilityTooltip(props: {
   }
 }) {
   const { t } = useTranslation()
-  const description = t(`ability_description.${props.ability}`)
+  const description = translateAbilityDescription(t, props.ability)
```
sylvainpolletvillard: no need to pass t as param, you can directly import it from i18next

app/config/game/abilities.ts
```diff
@@ -1,5 +1,37 @@
 import { Ability } from "../../types/enum/Ability"
 
+export type AbilityConfigValue = number | readonly number[]
+export type AbilityConfig = Readonly<Record<string, AbilityConfigValue>>
+
+export const AbilityConfigs = {
+  [Ability.AQUA_RING]: {
+    heal: [20, 40, 80, 160]
```
sylvainpolletvillard: would need a way to store the ap/luck scaling info for these numbers. I suggest using the string format convention we already use in translations

suggestion: 

```suggestion
    heal: "[20,40,80,160,SP]"
```

then the computeValue function would take care of parsing this format

app/core/abilities/aqua-ring.ts
```diff
@@ -1,12 +1,16 @@
+import { AbilityConfigs } from "../../config/game/abilities"
+import { Ability } from "../../types/enum/Ability"
 import type { Board } from "../board"
 import type { PokemonEntity } from "../pokemon-entity"
 import { AbilityStrategy } from "./ability-strategy"
 
+const abilityConfig = AbilityConfigs[Ability.AQUA_RING]
+
 export class AquaRingStrategy extends AbilityStrategy {
   requiresTarget = false
```
sylvainpolletvillard, line 8: suggestion: make the config a field of the AbilityStrategy

```suggestion
  requiresTarget = false
  config = AbilityConfigs[Ability.AQUA_RING]
```

then you can refer it with
```
const heal = this.computeValue(this.config.heal, pokemon)
```
or either
```
const heal = this.computeValue("heal", pokemon)
```

app/public/src/pages/utils/ability-description.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const INTERPOLATION_VARIABLE_REGEXP = /{{\s*([^{},\s]+)\s*}}/g
+const DYNAMIC_VALUE_FORMAT_REGEXP = /\[[^\]\n]*{{\s*[^{}]+\s*}}[^\]\n]*\]/g
+
+export function translateAbilityDescription(
+  t: TFunction,
+  ability: Ability
+): string {
+  const abilityConfig = getAbilityConfig(ability)
+  return abilityConfig
```
sylvainpolletvillard: no need for ternary condition, passing undefined is fine

app/public/src/pages/utils/ability-description.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export function resolveDescriptionPreview(
+  path: string,
+  template: string
+): string {
+  const abilityConfig = getConfigFromDescriptionPath(path)
+  if (!abilityConfig) return template
+
+  return template.replace(
```
sylvainpolletvillard: not sure why you need that regex stuff here, why not reusing translateAbilityDescription for preview ?

app/public/src/pages/utils/ability-description.ts
```diff
@@ -0,0 +1,80 @@
...
+    )
+  ].sort()
+  if (expected.join("|") !== actual.join("|")) {
+    return `Placeholders must match English: ${expected.map((name) => `{{${name}}}`).join(", ")}`
+  }
+
+  const expectedFormats = getDynamicValueFormats(englishTemplate)
+  const actualFormats = getDynamicValueFormats(targetTemplate)
+  if (expectedFormats.join("|") !== actualFormats.join("|")) {
+    return `Dynamic value formatting must match English (including SP/LK): ${expectedFormats.join(", ")}`
```
sylvainpolletvillard: if those formats per rank are part of the ability config and no longer the translation, no need to validate anything here

sylvainpolletvillard (changes requested):

app/config/game/abilities.ts
Cinnlite: on hold: Will remove the different AP Scaling factors from abilities in separate PR and continue with this afterwards.

app/core/abilities/aqua-ring.ts
Cinnlite: done for all abilities touched so far

app/public/src/pages/utils/ability-description.ts
Cinnlite: you are right, replaced it

sylvainpolletvillard:
Yes that's better. Let's make sure we get to the best code ergonomics before having to apply this on 500+ abilities

here's a suggestion alt method to compute scaled config for several fields at once:

```
const { damage, freezeChance } = this.computeConfigWithScaling(pokemon)
```

equivalent to

```
const damage = this.computeValue(this.config.damage, pokemon)
const freezeChance = this.computeValue(this.config.freezeChance, pokemon
```

Cinnlite:
> Yes that's better. Let's make sure we get to the best code ergonomics before having to apply this on 500+ abilities
> 
> here's a suggestion alt method to compute scaled config for several fields at once:
> 
> ```
> const { damage, freezeChance } = this.computeConfigWithScaling(pokemon)
> ```
> 
> equivalent to
> 
> ```
> const damage = this.computeValue(this.config.damage, pokemon)
> const freezeChance = this.computeValue(this.config.freezeChance, pokemon
> ```

I implemented computeConfigWithScaling and updated the migrated abilities to use it (added 4 more abilities). While reviewing I noticed that moving AP scaling out of handleSpecialDamage may affect some edge cases like Magic Bounce. I’ll need to review each ability against its previous behavior so we don’t introduce unintended gameplay changes. I'm worried about letting AI do all 500-600 abilities we have and reviewing later, cause I might miss something but we'll see how it goes.

sylvainpolletvillard:
I'm worried too. I used AI for adding fourth tier of damage for star piece and it got many things wrong. It will need to be a very precise prompt with lots of supervising

sylvainpolletvillard:
just finished reading the whole plan, looking good 👍 
The whole plan can be attached to the AI agent context, that will certainly help

here's my take on the questions at the end:

> Should the migration remain in one large PR with separate commits, or be split into category PRs?

I suggest splitting PR following the steps you described at Migration order

> Should patch notes permanently keep their isolated legacy number formatting?

yes patchnotes should remain static text, things change too much over time

> Is the proposed scope too much and you think we should keep the scope smaller?

I think it's ambitious but the proposed migration order makes it more progressive and therefore more approachable

PR #4045, opened 2026-07-26
Fix spectators loading a game with no map or music

brobinett:
Spectating a live game sometimes loads with a blank map and no music. Reloading fixes it.

The map tilesets and region music are queued in LoadingManager.preload only after preloadMaps's fetch resolves, so they can be added to the loader after its initial run has already completed. Phaser's loader only joins a file to the active run while it's loading, so those late assets never load. The loader's first complete event still fires, the client reports LOADING_COMPLETE, and for a spectator the server echoes it back immediately, so startGame runs before the map and music are ready.

Fix: Report LOADING_COMPLETE only after LoadingManager has finished its full preload

### DIFF
--- a/app/public/src/game/components/loading-manager.ts
+++ b/app/public/src/game/components/loading-manager.ts
@@ -19,6 +19,7 @@ export default class LoadingManager {
   scene: Phaser.Scene
   loadingBar: GameObjects.Container | null = null
   statusMessage: string
+  ready: Promise<void>
 
   constructor(scene: Phaser.Scene) {
     this.scene = scene
@@ -32,7 +33,7 @@ export default class LoadingManager {
       this.statusMessage = t("loading_complete")
     })
 
-    this.preload()
+    this.ready = this.preload()
   }
 
   async preload() {
@@ -111,6 +112,13 @@ export default class LoadingManager {
 
     // load missingno as default pokemon texture if not found
     loadCompressedAtlas(scene, "0000")
+
+    if (scene instanceof GameScene) {
+      await new Promise<void>((resolve) => {
+        scene.load.once("complete", () => resolve())
+        scene.load.start()
+      })
+    }
   }
 }
 
--- a/app/public/src/game/components/loading-manager.ts
+++ b/app/public/src/game/components/loading-manager.ts
@@ -19,7 +19,7 @@ export default class LoadingManager {
   scene: Phaser.Scene
   loadingBar: GameObjects.Container | null = null
   statusMessage: string
-  ready: Promise<void>
+  preloadingPromise: Promise<void>
 
   constructor(scene: Phaser.Scene) {
     this.scene = scene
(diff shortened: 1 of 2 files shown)


app/public/src/game/scenes/game-scene.ts
```diff
@@ -103,18 +103,20 @@ export default class GameScene extends Scene {
 
   preload() {
     resetSpriteCounts()
-    this.loadingManager = new LoadingManager(this)
+    const loadingManager = (this.loadingManager = new LoadingManager(this))
```
sylvainpolletvillard: that constant is unecessary and double affectation in one line is not readable

app/public/src/game/components/loading-manager.ts
```diff
@@ -19,6 +19,7 @@ export default class LoadingManager {
   scene: Phaser.Scene
   loadingBar: GameObjects.Container | null = null
   statusMessage: string
+  ready: Promise<void>
```
sylvainpolletvillard: suggestion to rename to preloadingPromise. ready sounds like a boolean

app/public/src/game/components/loading-manager.ts
```diff
@@ -111,6 +112,13 @@ export default class LoadingManager {
 
     // load missingno as default pokemon texture if not found
     loadCompressedAtlas(scene, "0000")
+
+    if (scene instanceof GameScene) {
+      await new Promise<void>((resolve) => {
+        scene.load.once("complete", () => resolve())
```
sylvainpolletvillard, line 119: i suggest adding a comment to make things more clear

```suggestion
        // start another Phaser loading queue after the fetch requests of preloadMaps have been awaited
        scene.load.once("complete", () => resolve())
```

sylvainpolletvillard (changes requested):
I wish we could find a way to make the whole map preloading process using a single Phaser loading queue, but I don't know how. For now this is a good enough workaround starting another queue after all the await in preload()

app/public/src/game/scenes/game-scene.ts
brobinett: Done

app/public/src/game/components/loading-manager.ts
brobinett: Done

app/public/src/game/components/loading-manager.ts
brobinett: Done

brobinett:
> I wish we could find a way to make the whole map preloading process using a single Phaser loading queue, but I don't know how. For now this is a good enough workaround starting another queue after all the await in preload()

Agreed, a single queue would be nicer. The per-map tileset list only comes back from the async tilemap fetch, so those load.image calls can't be issued until after the first run may have already drained. I think the way to fold it into one queue would be to prefetch the tilemap json before preload.

sylvainpolletvillard:
I believe the Phaser Loader API can be used to add our own Promise in the queue, but I didn't take the time to dig into it. Anyway, merging as it is, thanks for the help !

PR #4051, opened 2026-07-28
Fix spectators joining a game with no player selected

brobinett:
Joining a game as a spectator from the lobby shows a board, but no player is actually selected. Clicking any avatar fixes it.

GameContainer only learns it is spectating in initializeSpectactor, which runs from spectators.onAdd, after the players.onAdd loop has already run initializePlayer with this.spectate still false. Its spectator branch (this.spectate && !this.player) is not taken for a lobby spectator and setPlayer is never called, so gameContainer.player stays undefined and playerIdSpectated stays empty (replays set spectator in the constructor and do use that branch). A board still appears, because startGame falls back to playerUids[0], but everything keyed off the spectated player does not.

Fix: select a player in initializeSpectactor the same way playerClick does. The selection has to happen before initializeGame. initializeGame now passes spectatedPlayerId into the scene, and startGame reads it in init.

The player is picked with sortPlayersByRankAndTeam over the players still alive, same order as player list.

sortPlayersByRankAndTeam is unchanged apart from being made generic, so it returns Player to this caller instead of IPlayer; game-players.tsx is unaffected.

### DIFF
--- a/app/public/src/game/game-container.ts
+++ b/app/public/src/game/game-container.ts
@@ -655,6 +655,7 @@ class GameContainer {
               this.room.state.gameMode
             )[0] ?? players[0]
           if (spectated) {
+            this.room.send(Transfer.SPECTATE, spectated.id)
             this.setPlayer(spectated)
             const simulation = this.room.state.simulations.get(
               spectated.simulationId
--- a/app/public/src/game/game-container.ts
+++ b/app/public/src/game/game-container.ts
@@ -649,16 +649,16 @@ class GameContainer {
       if (this.room.state.players.size > 0) {
         if (!this.player) {
           const players = schemaValues(this.room.state.players)
-          const spectated =
+          const playerToSpectate =
             sortPlayersByRankAndTeam(
               players.filter((p) => p.alive),
               this.room.state.gameMode
             )[0] ?? players[0]
-          if (spectated) {
-            this.room.send(Transfer.SPECTATE, spectated.id)
-            this.setPlayer(spectated)
+          if (playerToSpectate) {
+            this.room.send(Transfer.SPECTATE, playerToSpectate.id)
+            this.setPlayer(playerToSpectate)
             const simulation = this.room.state.simulations.get(
-              spectated.simulationId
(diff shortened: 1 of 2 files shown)


app/public/src/game/game-container.ts
```diff
@@ -315,7 +316,8 @@ class GameContainer {
     this.game.scene.start("gameScene", {
       room: this.room,
       uid: this.uid,
-      spectate: this.spectate
+      spectate: this.spectate,
+      spectatedPlayerId: this.player?.id
```
sylvainpolletvillard, line 320: this new prop doesn't seem to be used by GameScene

app/public/src/game/game-container.ts
```diff
@@ -645,6 +647,24 @@ class GameContainer {
     if (this.uid === uid) {
       this.spectate = true
       if (this.room.state.players.size > 0) {
+        if (!this.player) {
+          const players = schemaValues(this.room.state.players)
+          const spectated =
```
sylvainpolletvillard: suggestion: rename spectated to playerToSpectate

sylvainpolletvillard (changes requested):
looking good, just minor comments

app/public/src/game/game-container.ts
brobinett: It's used by the replay system for the seek re-attach path. `GameScene` reads it in `startGame()`:
```ts
const player = this.room.state.players.get(
  this.spectate ? (this.spectatedPlayerId ?? playerUids[0]) : this.uid
) as Player
this.setMap(player.map)
```
It's needed because `setPlayer()` can't set the map at that point: both paths into `initializeGame()` call `setPlayer()` while `this.game` is still null, so `setPlayer`'s `this.gameScene?.setMap(player.map)` no-ops. Without the prop the scene falls back to `playerUids[0]`, so a spectator gets the map of the first player in `room.state.players` instead of the player they're actually watching.

app/public/src/game/game-container.ts
brobinett: Done

PR #4052, opened 2026-07-29
Load portraits on demand instead of preloading every pokemon index

brobinett:
`LoadingManager.preload()` queues one `scene.load.image` per `PkmIndex`, ~1178 individual portrait files, on every game load and every refresh, and the cache-first service worker turns each one into a separate Cache Storage write. That is much slower in Firefox than in Chrome, so Firefox players sit on the loading screen for a long time, and because game start waits for the slowest client a single Firefox player holds up the whole lobby. The `portrait-<index>` texture has only two consumers, the damage number in `battle-manager` and the shard gain in `wanderers-manager`, so this loads portraits on first use instead, mirroring what `lazyLoadAnimations` already does for pokemon atlases. Each popup renders its text immediately and adds the portrait once the texture arrives.

### DIFF
--- a/app/public/src/game/components/battle-manager.ts
+++ b/app/public/src/game/components/battle-manager.ts
@@ -37,6 +37,7 @@ import type AnimationManager from "../animation-manager"
 import { DEPTH } from "../depths"
 import type GameScene from "../scenes/game-scene"
 import { displayAbility, displayHit } from "./abilities-animations"
+import { loadPortrait } from "./loading-manager"
 import PokemonSprite from "./pokemon"
 import {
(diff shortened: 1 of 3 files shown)


sylvainpolletvillard:
You missed the most important use, in GamePokemonPortrait (see getCachedPortrait). It ensures the image in the shop is displayed immediately, which is crucial when rolling down fast for the gameplay.

brobinett:
Thanks for catching it, I tried warming the cache but I think this approach will never be good enough. Not worth a gameplay regression for performance at load time.

PR #4053, opened 2026-07-29
Fix grammar/typo in english quest reward message

ninesl:
Fixed a typo `here are your reward` is incorrect grammar. Sorry Chatot :(


ninesl:
Could also be `"Here is your reward:"`

PR #4083, opened 2026-08-23
Add board effects to the Battle Stats

tegberen:
## Add board effects to the Battle Stats

Board effects and certain weather now track their own damage in Battle Stats instead of being ignored or credited to Pokémon, and are also documented in the Wiki under Glossary Tab.

The change is built on four points:

1. **Its own row in the Battle Stats.** Ember, Hail, Spikes, Stealth Rocks, Poison Gas, Toxic Spikes, Sandstorm, Lightning Strike, Tidal Wave and Curse each get a row in the damage meter, shown with the effect's icon in place of a portrait. Tidal Wave also gets its healing on the Heal and Shield tab.
2. **Damage numbers you can read.** Damage from a board effect now floats up with that effect's icon, coloured to the damage type. Before, it showed no number at all, because there was no attacker to draw.
3. **A Board effects section in the wiki glossary.** All fourteen board effects are listed with their icon and a one-line explanation, and the ones that show up in the Battle Stats carry a tag saying so. Ability and passive descriptions no longer repeat those mechanics — they simply name the effect, which renders as an icon.
4. **Two attribution fixes.** Poison from Poison Gas or Toxic Spikes, and burn from an Ember tile, kept being credited to whoever poisoned or burned the target first, instead of to the tile.

Everything is additive: no ability, item or synergy changes were made.

### Reviewing this PR (suggested reading order)

* `app/core/simulation.ts` — `creditBoardEffectDamage`, the one helper that records this damage, plus the Storm lightning and Tidal Wave call sites.
* `app/models/colyseus-models/status.ts` — the poison, burn and curse ticks, and the two ownership fixes in `triggerPoison` and `triggerBurn`.
* `app/core/pokemon-state.ts` — sandstorm and the four damaging tiles.
* `app/types/index.ts` and `app/types/enum/Effect.ts` — the two short lists that drive everything else: which effects have a row, and which are documented.

The rest is presentational: the meter components, the floating damage icon, the wiki section, and the locale strings.

### How it works

**Recording the damage.** Board effect damage has no attacker, so nothing was recording it. `creditBoardEffectDamage` adds it to a row on the meter of the side *opposing* the Pokémon taking the damage, so it reads exactly like damage from a real attacker: what your board effects do to the enemy shows in your column, and what theirs do to your team shows in theirs. The same helper sends the floating damage number, since the normal damage path skips it when there is no attacker.

**Where the rows appear.** A row only shows once it has something to display, and never on the Damage blocked tab, since a tile cannot take damage. These rows are also excluded from the end-of-game personal bests (highest damage by a single unit), because they are team-wide totals rather than a Pokémon.

**The two fixes.** When a tile poisons or burns you it does so without an owner, but the status only ever remembered the last owner it was given and never cleared it. So a Pokémon that poisoned or burned you earlier kept the credit for every later tick — and since the tile refreshes the status every second, that could last the whole fight. Now the tile takes ownership once it is the thing keeping the status alive.

**Descriptions.** Sixteen ability descriptions and four passive descriptions were reworded (in english). Each one that creates a board effect now begins with `BOARD_EFFECT:` and names the effect instead of restating how it behaves, which removes a lot of repeated text and keeps the rules in one place.

### New files

* `app/public/src/pages/component/game/board-effect-dps.ts` — maps each board effect row to its icon and name.
* `app/public/src/assets/icons/board_effects/` — 14 new icons.

### Changes to existing files

* `simulation.ts`, `pokemon-state.ts`, `status.ts` — record the damage at each place a board effect deals it.
* `player.ts` — keeps these rows out of the per-Pokémon records.
* `battle-manager.ts`, `loading-manager.ts`, `game-container.ts` — the floating icon and its preloading.
* `game-dps.tsx`, `game-dps-heal.tsx` and the three meter lists — draw the icon, and hide rows with nothing in them.
* `descriptions.tsx` — renders a board effect name as an icon wherever it appears.
* `wiki-glossary.tsx` / `.css` — the new section, its note and the Battle Stats tag.
* `replay-event-format.ts` — names the board effect in moonpanda's replay event log 
* `Effect.ts`, `types/index.ts` — the lists of documented effects and of the rows they map to.
* `translation.json` — new effect names and descriptions, the glossary text, and the reworded ability and passive descriptions.

### Notes

* English only. New strings live under `effect.*`, `effect_description.*` and `wiki.glossary.*`.


sylvainpolletvillard:
Thanks a lot @tegberen <3

PR #4085, opened 2026-08-23
Don't route portraits through the service worker cache

brobinett:
My second attempt at solving the Firefox cache storage problem. (first attempt: #4052)

`sw.js` sends every `/assets/` request through `cacheFirst`, so each of the ~1200 portrait files a game load asks for becomes a Cache Storage operation. Portraits do not need that cache. The set is ~1260 files and ~730 KB, and it is already served with a long `max-age`, so the browser's own http cache covers it.

`preloadPortraits` untouched, all portraits still load before the game starts, so `getCachedPortrait` still returns base64 on the first render.

Measured on master with only `sw.js` differing, two cold runs each, from the spectate click until the loading screen goes:

| | cold load | portrait requests | Cache Storage entries |
|---|---|---|---|
| firefox, master | 70.3s / 75.7s | ~1186 | 1176 |
| firefox, this | 7.3s / 7.5s | ~1188 | 0 |
| chromium, master | 9.0s / 8.1s | ~1189 | 1176 |
| chromium, this | 7.4s / 7.9s | ~1189 | 0 |

Tradeoffs:

This swaps a Cache Storage read for the network-stack fallback, which Chrome does more slowly than it reads Cache Storage. In practice the difference is negligible, a fraction of a second.

Cache Storage is per-origin and quota-managed; the browser http cache is a global LRU that unrelated browsing evicts. A player whose http cache has dropped the portraits has to re-request them. 

"Delete cache" in Options > Game Files stops being a way to force portraits to refetch. These rarely change, so I consider this to be acceptable. Options > Game Files is otherwise unaffected.

The first commit serves `/assets/portraits` with a cache lifetime, because production only gets one from Cloudflare and a self-hosted instance would otherwise trade Cache Storage hits for ~1200 revalidations per load. If you would rather leave caching to the CDN, it's cuttable; the second commit is fine on its own.

### DIFF
--- a/app/app.config.ts
+++ b/app/app.config.ts
@@ -278,6 +278,13 @@ export const server = defineServer({
 
     app.use(cors())
     app.use(express.json())
+    // portraits use the browser's http cache instead of the service worker
+    app.use(
+      "/assets/portraits",
+      express.static(path.join(clientSrc, "assets", "portraits"), {
+        maxAge: isDevelopment ? 0 : "31d"
+      })
+    )
     app.use(express.static(clientSrc))
 
     app.get("/", (req, res) => {
--- a/app/public/dist/client/sw.js
+++ b/app/public/dist/client/sw.js
@@ -42,7 +42,8 @@ self.addEventListener("fetch", async (event) => {
   const url = event.request.url
   if (
     event.request.method === "GET" &&
-    (url.includes("/assets/") || url.includes("/SpriteCollab/"))
+    (url.includes("/assets/") || url.includes("/SpriteCollab/")) &&
+    !url.includes("/assets/portraits/") // portraits are left to the browser's http cache
   )
     cacheFirst(event)
 })


sylvainpolletvillard:
So the performance problem on Firefox did not come from HTTP Requests but from the service worker with Cache API ? It is quite surprising, is this problem documented somewhere ?

The major tradeoff is the increase of bandwidth costs. I can't measure how that change will increase our costs, how many players use private browsing or have automatic cache clearing policies.

brobinett:
I don't know how well known the issue is, but it's easy to reproduce if you want to see for yourself. You can run this in the browser console on any site. I get 0.98 ms/op in Chromium and 47.9 ms/op in Firefox. It's the writes specifically that are the issue, reads are much closer.
```js
const N = 1000
const cache = await caches.open("perftest")
const body = new Uint8Array(600)
const t = performance.now()
for (let i = 0; i < N; i++) await cache.put(new Request(`/perftest/${i}`), new Response(body))
console.log((performance.now() - t) / N, "ms/op")
await caches.delete("perftest")
```

brobinett:
On bandwidth, I think it shouldn't increase costs, but I'm not sure I can say for sure.

`cacheFirst` falls back to `fetch(event.request)`, which is an ordinary fetch and goes through the browser's http cache. Portraits already come back `public, max-age=2678400` from Cloudflare, so the http cache holds the whole set for 31 days, the Cache Storage copy is sitting in front of a cache that already has everything.

I'll post with some numbers when I have them.

ql1111:
even with max-age=2678400 i dont think it will increase bandwith  cost as cloudflare return a 304 Not-Modified HTTP if the ressource remain unchanged. Even at the end of the 31 days.

see https://cloudflare-docs.cloudflare-docs.workers.dev/cache/reference/etag-headers/#strong-etags#1

sylvainpolletvillard:
right Cloudflare can take the hit for us. OK for me, waiting for Keldaan's opinion since he's the one paying the bills.

Maxouille64:
is it me or firefox 154 fixed the issue? game load way faster on both production and local server. I dont feel any difference between chrome and firefox

brobinett:
I'm seeing not fixed on 154 or 155. 154 shipped two days after the last PAC cache wipe though, so you've probably been on a warm cache ever since, and this is only an issue for cold loads. Try Options > Game Files > Delete cache, then reload. (A private window isn't a valid test, Firefox disables service workers there so it's fast for the wrong reason.)

PR #4086, opened 2026-08-23
fix nullify bandanna duplication

ql1111:
how to recreate the issue: start a game with a scarf given by either wooloo or the shop then reach normal 3 synergy and create a nullify bandanna with both scarves. Then reach normal 5 and you get another nullify bandana instead of a silk scarf.

what this fix does:

- create a player.extraScarves counter that increases every time the player is given a scarf in the shop or with the fur coat ability
- change the player.getScarvesItemsWithNbScarves logic to give the correct weight of 1 instead of 2 when the nullify bandana is made of one normal synergy's scarf and one extra scarf.

### DIFF
--- a/app/models/colyseus-models/player.ts
+++ b/app/models/colyseus-models/player.ts
@@ -497,12 +497,12 @@ export default class Player extends Schema implements IPlayer {
   }
 
   getScarvesItemsWithNbScarves(n: number): Item[] {
+    n=n+this.extraScarves
     let i = 0
     const scarves: Item[] = []
     while (n > 0) {
       const scarf = this.scarvesItems[i] ?? Item.SILK_SCARF
-      const nbOfNullifyBandannas = scarves.filter(item => item === Item.NULLIFY_BANDANNA).length;
-      n -= ((scarf === Item.NULLIFY_BANDANNA) && ( (this.extraScarves % 2 === 0) || (nbOfNullifyBandannas > 0) )) ? 2 : 1
+      n -= scarf === Item.NULLIFY_BANDANNA ? 2 : 1
       if (n >= 0) {
         scarves.push(scarf)
         i++
@@ -525,7 +525,7 @@ export default class Player extends Schema implements IPlayer {
     )
     const newNbNormalScarves = getSynergyTier(updatedSynergies, Synergy.NORMAL)
     const newScarves = this.getScarvesItemsWithNbScarves(newNbNormalScarves)
-
+      
     if (newScarves.length > previousScarves.length) {
       // some scarves are gained
       const gainedScarves = newScarves.slice(
--- a/app/models/colyseus-models/player.ts
+++ b/app/models/colyseus-models/player.ts
@@ -198,6 +198,7 @@ export default class Player extends Schema implements IPlayer {
   specialGameRule: SpecialGameRule | null = null // its easier to duplicate this here and in gamestate than passing gamestate everywhere we need it
   shopsSinceLastUnownShop: number = 0
   regions: DungeonPMDO[] = []
+  extraScarves: number = 0
   unownReminiscences: number = 0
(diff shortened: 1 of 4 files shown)


app/models/colyseus-models/player.ts
```diff
@@ -500,7 +501,8 @@ export default class Player extends Schema implements IPlayer {
     const scarves: Item[] = []
     while (n > 0) {
       const scarf = this.scarvesItems[i] ?? Item.SILK_SCARF
-      n -= scarf === Item.NULLIFY_BANDANNA ? 2 : 1
+      const nbOfNullifyBandannas = scarves.filter(item => item === Item.NULLIFY_BANDANNA).length;
+      n -= ((scarf === Item.NULLIFY_BANDANNA) && ( (this.extraScarves % 2 === 0) || (nbOfNullifyBandannas > 0) )) ? 2 : 1
```
sylvainpolletvillard: just to be sure because i don't fully understand the logic here: does that cover the case where a player makes a nullify bandanna with 2 extra scarves then activate normal 3 ? And also, could this be simplified by replacing `n` by `n + this.extraScarves` ?

ql1111: yes it covers that case.

In theory doing simply n + this.extraScarves should fix the whole thing, its the first thing i tried but for some reason it didnt worked. Getting one extra scarf and immediately crafting 1 silk scarf item then getting normal 2 and crafting a nullify bandanna broke the thing.

i'll dig into it

another solution would be to save the cost of every scarf items in term of silk scarves from the normal synergy. in an Array. But it would be overwhelming, maybe.

sylvainpolletvillard:
If you're confident about that solution we can go for it. But just so you know, I used to compute extra scarves in the past, exactly like you did here, and have iterated several times on numerous bugs as well. Maybe just for extra safety I should list back all the bugs about scarves to make an exhaustive lists of tests

sylvainpolletvillard:
test A: mix between normal scarf and extra scarf

- get 1 extra scarf and 1 scarf from normal 3
- craft a nullify bandanna
- remove normal 3: you should lose nullify bandanna
- readd normal 3: you should get back your nullify bandanna
- activate normal 5: you should get another silk scarf

test B: bandanna with 2 normal scarves overlapping 2 levels of the synergy

- get normal 5 and 2 scarves
- craft a nullify bandanna
- get from normal 5 to normal 3: you should lose nullify bandanna
- remove normal 3, re-add normal 3: you should gain no scarf
- re-add normal 5: you should get back your nullify bandanna

test C: nullify bandanna crafted with extra only

- get 2 extra scarves and 2 normal scarves from normal 5
- craft 2 nullify bandanna
- get from normal 5 to normal 3: you should lose one nullify bandanna
- remove normal 3, re-add normal 3: you should gain no scarf
- re-add normal 5: you should get back your second nullify bandanna
- activate normal 7: you should get another silk scarf

Related bug reports:
- https://discord.com/channels/737230355039387749/1539117018417078273
- https://discord.com/channels/737230355039387749/1487822549504102440
- https://discord.com/channels/737230355039387749/1464963053421334678

app/models/colyseus-models/player.ts
ql1111: ok so I found why doing n=n+1 didnt worked and found a cleaner fix to it. testing the tests bellow rn.

ql1111:
I did all the 3 tests and additional tests with multiple recycle tickets to match as much case as possible and so far I didnt had any issue. 

The trick was to remove this condition `if (player.scarvesItems.length < nbScarvesBasedOnNormalSynergy) {player.scarvesItems.push(result)}`  when trying to add freshly built scarf item to the scarveItems array. Because now thoses scarve items are handled by the extraScarves counter.

sylvainpolletvillard:
Excellent, that logic was quite hard, thanks a lot !

PR #4091, opened 2026-08-24
Add a FPS limiter

RobCubed:
My laptop is a bit old and tries to light itself on fire over the course of a single PAC round. Did some profiling and found that simply limiting the FPS brings it from being on fire to merely smoldering, which is a significant improvement.

Tried to keep the UI to established elements since I'm new to the codebase.


ql1111:
Wouldn't limiting to 60hz be better for older devices and tablet/chromebooks? Those are usually C a pped at 60hz. 

Very good idea though, I'll do some measurements.

RobCubed:
> Wouldn't limiting to 60hz be better for older devices and tablet/chromebooks? Those are usually C a pped at 60hz.
> 
> Very good idea though, I'll do some measurements.

It's adjustable in the settings - I just set 144 as the default to largely mimic the existing behavior. I was actually using 10-15 FPS on my laptop while testing to keep the CPU not at 100%

sylvainpolletvillard:
@RobCubed Thanks for your contribution. Can you add a no limit option, just like it is currently ? The "0" value can be used for that, see https://docs.phaser.io/api-documentation/typedef/types-core#fpsconfig

RobCubed:
> @RobCubed Thanks for your contribution. Can you add a no limit option, just like it is currently ? The "0" value can be used for that, see https://docs.phaser.io/api-documentation/typedef/types-core#fpsconfig

Done

sylvainpolletvillard:
Good to merge for me 👍

PR #4126, opened 2026-09-16
Re-publish room listings that go missing, and stop the lobby cron deleting live ones

brobinett:
Two player-facing issues here, caused by the same root issue. I also fixed some issues in the stale room cron.

Player-facing bugs:
Lobbies get hidden, and trying to join that game mode creates a new lobby of that type. A player leaving the hidden lobby makes it visible again. I'm pretty sure this is also the bug that hides tournament lobbies.

In progress games become unrejoinable. A player who refreshes sees "room not found".

The issue:
A room and its listing have separate lifetimes. When a process misses a single IPC health-check ping, `@colyseus/core`'s `MatchMaker.healthCheckProcessId` calls `removeRoomsByProcessId` and `RedisDriver.cleanup` `HDEL`s the `roomcaches` entry of every room on that process. The rooms keep running; only their listings are gone, and nothing ever puts them back. `CustomLobbyRoom`'s cron then broadcasts `REMOVE_ROOM` for rooms it can no longer find in `matchMaker.query`; `JoinOrOpenRoomCommand` dedups against that same list, so it misses and opens a duplicate; and `joinById` and `MatchMaker.reconnect` both start with `driver.findOne({ roomId })`, so rejoining fails. Any join or leave blind-`HSET`s the listing back, which is why leaving and rejoining is the workaround. I reproduced it on two processes with Redis by stalling one of them.

I think this is also behind #3304 (lobby duplicated, duplicating its cron jobs with it) and #2859 (`NextTournamentStageCommand` called twice).

The stale room cron also had a couple independent bugs. It `forEach`es `this.rooms` while `removeRoom` splices that same array, so every removal skips the next element; and `matchMaker.query()` is a snapshot, so a listing re-persisted between the snapshot and the diff is `HDEL`ed despite being live.

The fix:
Since this is upstream Colyseus behavior, the best we can do is recovery, not prevention. The issues I noted at the top will still happen, they'll just be bounded to the length of the interval instead of requiring someone to leave/join to recover.

Rooms re-publish a listing that has gone missing. `PreparationRoom` and `GameRoom` check every 30 s whether their own listing still exists and re-publish it if not. The check is `matchMaker.driver.has(roomId)`, a single `HEXISTS`. The cost is one read per room per interval, it only writes when the listing is gone. The preparation room stops checking in `onGameStart`, so it never fights the cron's `gameStartedAt` clean-up. The lobby room does the same at 10s, so we get two attempts within the two 10-second retries #3304 added to `checkLobby`.

The stale-room cron no longer deletes live listings or skips rooms. The loop now iterates a copy, removes each room at most once per tick, and only `HDEL`s on the `gameStartedAt` branch.


app/rooms/custom-lobby-room.ts
```diff
@@ -625,32 +639,46 @@ export default class CustomLobbyRoom extends Room {
...
+
           query.forEach((data) => {
             if (!this.rooms?.map((r) => r.roomId).includes(data.roomId)) {
               // if the query room was not in this.rooms, add it
               this.addRoom(data.roomId, data)
             }
           })
-          this.rooms?.forEach(async (room, roomIndex) => {
+
+          // iterate backwards: removeRoom splices this.rooms, so a forward loop skips elements
```
sylvainpolletvillard: why not just forEach like before to not have this problem of indexed loop and splice ?

sylvainpolletvillard:
>  It forEaches this.rooms while removeRoom splices that same array, so every removal skips the next element; 

forEach is precisely supposed to avoid this kind of problem since it doesn't rely on indexes

app/rooms/custom-lobby-room.ts
```diff
@@ -625,32 +639,46 @@ export default class CustomLobbyRoom extends Room {
...
-            ) {
-              this.presence.hdel("roomcaches", room.roomId)
+            const startedBefore = (ms: number) =>
+              gameStartedAt != null &&
+              new Date(gameStartedAt).getTime() < Date.now() - ms
+
+            if (!listedRoomIds.has(room.roomId)) {
+              // no hdel: query is a snapshot, and a room re-listed since is live
               this.removeRoom(roomIndex, room.roomId)
+              continue // one removal per room per tick
```
sylvainpolletvillard: this can be replaced with an "else" on next block, or an early return if going back to forEach

sylvainpolletvillard:
OK for the rest, thank you !

There's a discussion to have on Colyseus side because this should definitely not be something we have to do ourselves

app/rooms/custom-lobby-room.ts
brobinett: done

app/rooms/custom-lobby-room.ts
brobinett: changed to forEach over a copy

brobinett:
> > It forEaches this.rooms while removeRoom splices that same array, so every removal skips the next element;
> 
> forEach is precisely supposed to avoid this kind of problem since it doesn't rely on indexes

I think that forEach does rely on indexes.
```ts
const a = [1, 2, 3, 4]
a.forEach((x, i) => a.splice(i, 1))
console.log(a) // [2, 4]
```
But taking another look at this I do think there's another approach that's better, because the real problem is splicing the array that's being iterated. Changed to forEach over a copy with the index looked up per room the same way that subscribeLobby does.

sylvainpolletvillard:
Oh I learnt something new today. I was convinced it did not have this issue, I don't know why I did not realize it after all this time. Probably because i use filter over splice most of the time. Good to merge for me then !

sylvainpolletvillard:
Good to merge 👍

PR #4128, opened 2026-09-20
Dispose duplicate game and tournament rooms left by a create timeout

brobinett:
A ranked match sometimes shows up twice in the in-game list. The empty copy plays itself out and changes every player's ELO a second time. In tournaments, a bracket can get two lobbies, and the empty one eliminates the whole bracket after 10 minutes while its players are still in their game. With this fix, the copy removes itself within about 3s of appearing.

The bug:
`matchMaker.createRoom` sends the create request to another process and waits 2s. On a timeout it creates the room locally, but the request is never withdrawn. If the other process was only stalled, it creates a second room when it resumes. RANKED and TOURNAMENT rooms have `autoDispose` off, so the empty copy stays. A game copy force-starts at `MAX_LOADING_TIME` and writes ELO and game history for all 8 players in `onDispose`. A lobby copy's `autoStartTimeout` publishes `tournament-match-end` with no players, and `EndTournamentMatchCommand` eliminates everyone in the bracket.

Reproduced on two processes with Redis by stalling one of them, and checked on colyseus 0.17.51 and on #4125's 0.18.14.

#4126 comes from a stalled process too, by a different path. There the listings go missing, here the create request is answered twice.

The fix:
The three create calls now pass a random `creationId`, and the caller stores the returned room's id under that key in presence. Each of those rooms checks the key every 3 s for up to 3 minutes. If the key names another room, it disposes itself through `onRoomDeleted`, the same path as an admin delete. Both room types also refuse joins once deleted. The keys expire after 6 hours.

Tradeoffs:
Anyone who got into the copy before then sees "This room has been deleted" and goes back to the lobby, with no ELO change. A tournament player then has to find the real lobby. I consider this acceptable given how briefly the copy exists.


app/rooms/game-room.ts
```diff
@@ -720,6 +725,11 @@ export default class GameRoom extends Room<{ state: GameState }> {
         )
       }
 
+      if (this.deleted) {
```
sylvainpolletvillard, line 728: What is the purpose of the deleted flag ? If the room is deleted, the disconnect method has been called which prevents anyone from joining it anyway

app/services/duplicate-rooms.ts
```diff
@@ -0,0 +1,47 @@
+import { matchMaker } from "colyseus"
+import type GameRoom from "../rooms/game-room"
+import type PreparationRoom from "../rooms/preparation-room"
+import { logger } from "../utils/logger"
+
+const ROOM_CREATION = "room_creation"
+const ROOM_CREATION_TTL = 6 * 60 * 60 // a stall longer than this goes unchecked
+const DUPLICATE_CHECK_INTERVAL = 3000 // under CheckAutoStartRoom's 5 s
+const DUPLICATE_CHECK_TIMEOUT = 3 * 60 * 1000
```
sylvainpolletvillard, line 9: I think these constants should depend on the room type and be passed as arguments to createRoomWithDuplicateCheck

app/services/duplicate-rooms.ts
```diff
@@ -0,0 +1,47 @@
...
+  const check = room.clock.setInterval(async () => {
+    if (Date.now() - startedAt > DUPLICATE_CHECK_TIMEOUT) {
+      check.clear() // this last tick still reads
+    }
+    try {
+      const createdId = await room.presence.get(
+        `${ROOM_CREATION}:${creationId}`
+      )
+      if (!createdId) return // the creator may record the room after this copy exists
+      check.clear()
```
sylvainpolletvillard, line 38: I don't understand why the interval is clear immediately here

sylvainpolletvillard (commented):
Thanks, I think just like #4126 this should be reported as a bug on Colyseus repo. This will work as temporary workaround for now

